Issue #14, opened 2015-01-08
change R version in Depends to R 3.1.0


rmflight:
Does the package actually depend on having R 3.1.2? I was able to clone, change the Depends field in DESCRIPTION to R 3.1.0, and install on my system with R 3.1.1, and run it without difficulty.

jimhester:
No, it probably does not, `devtools::create()` creates packages with the current R as the Depends.  It is probably safe to change to 3.1.0.

Issue #13, opened 2015-01-08
summarize test coverage without coveralls?


rmflight:
Cool package! I noticed that if you run the `package_coverage` function, it gives you a ton of output, that I'm guessing gets uploaded to the *coveralls* service to give you a badge. However, when I'm trying to improve test coverage, I would want to be able to just re-run `package_coverage` as I add tests and see how the coverage changes.

Is there a way to generate a % coverage value from the `package_coverage` output? I'd be willing to write a `summarize` method if I knew how to interpret the values that are being returned.

jimhester:
if you run `as.data.frame(package_coverage())` it will give you a data.frame of the results, which shoud be easier to interpret.

However if all you are after is overall percentage coverage you can just run
```r
cov <- package_coverage()
sum(cov) / length(cov)
```

The value returned by `package_coverage()` is simply a named vector of counts per applicable line.

PR #16, opened 2015-01-08
Summary functions

rmflight:
Add two new functions useful for a package developer / maintainer on their own machine: 

* percent_coverage: gives how much test coverage there is by looking at how many zero values there is in the package_coverage output
* zero_coverage: returns just the zero locations, and the really relevant pieces of the package_coverage output to help the developer navigate to the file locations

Unfortunately, I can't find a way to test these on covr itself, I get a recursion issue if I do that.


jimhester:
Thanks for working on this!

You can add tests for your functions by running package_coverage on one of the test packages in `tests/testthat/` (note that test are run from the `tests/testthat` directory)

e.g.
```r
context("summary_functions")
cov <- package_coverage('TestS4')
test_that("percent_coverage returns correct output", {
expect_equals(percent_coverage(cov), .543)
})
```

Also an admittedly silly thing could you rename `n_lines` and `n_not_zero` to `num_lines` and `num_not_zero`.

Thanks again!

rmflight:
Do these work for you?

jimhester:
Yep that works.  I should probably make a project that has intermediate coverage so the tests are less trivial but we will leave that for another day.

rmflight:
Yes, ideally we would have that. I can probably create that package for testing the first of next week.

rmflight:
OK, so once I thought about it, that took about 10 minutes to do. 

This pull request now:

closes #13 
closes #14 
closes #17

jimhester:
Awesome, thanks for your work on this!  Looks great.

One thing that might be useful for these summary functions is to stratify the results by file (should be easy) or even by enclosing function (pretty challenging with the current covr implementation)

If you want to do either of the above in this pull I can wait to merge it or I can merge it as-is now and I can add the functionality later (or you can open a new pull request).

rmflight:
This might be an interesting option to add later, but I think I would merge as is for now.

My reasoning is that the `percent_coverage` gives a good overview of how the package is as a whole, and doesn't really need broken down by file. The `zero_coverage` has the `filename` column in the results, and this points a developer directly to the file and line that is not currently being tested. I was going to split out the results by file until I observed that the `coverage` data.frame is sorted by filename and line. This also goes for the enclosing function, because a developer would have the source right in front of them.

I guess one option would be to be able to specify which file one wants the coverage for, which in that case the easiest thing would be to break it up by file initially. If you want, open an issue and assign me to it?

jimhester:
Fair points, thanks for the contribution!

jimhester:
Also FYI if you put the 
```
closes #13 
closes #14 
closes #15 
closes #16 
closes #17
```
In your git commit message github will automatically close the issues when I merge the pull.

PR #30, opened 2015-01-12
fixed coverage of blocks containing just a name, e.g. {x}. add tests, am...

kforner:
- improved coverage for blocks with simple expressions, such as:
function(x) { x }
function() {
   x <- 1
  x
}

so now expressions such as your a() function:
a <- function(x) {
if (x <= 1) {
1
} else {
2
}
}
can be covered exactly (i.e the "1" and "2" expressions).

I also fixed the coverage statistics that assumed the unicity of first_line attributes. I added a test with a case with nested expressions which have same first_line.

I added some tests, fixed the legacy ones, and provided a Makefile that automates testing with the docker rocker/r-base.

In conclusion: with these modifications, your package give the same results that my patched R, results that were checked during a year, so that's really encouraging.

If you find this pull request interesting, we could discuss some further optimizations.


jimhester:
Thanks for this, overall it looks good, I will put some specific comments inline.

R/covr.R
```diff
@@ -16,8 +16,22 @@ trace_calls <- function (x, srcref = NULL) {
     lapply(y, trace_calls)
   }
 
+#  if (is.atomic(x) || is.name(x)) {
```
jimhester: Could you remove the commented lines here and at 61

R/covr.R
```diff
@@ -44,7 +58,10 @@ trace_calls <- function (x, srcref = NULL) {
       fun_body <- trace_calls(fun_body)
     }
 
-    formals(x) <- trace_calls(formals(x))
+    #formals(x) <- trace_calls(formals(x))
```
jimhester: What edge case causes the need for this extra logic?  Do you have a test for it?

R/summary_functions.R
```diff
@@ -14,13 +15,18 @@ percent_coverage <- function(coverage_result){
...
 #' any locations where there is \bold{0} test coverage.
 #'
-#' @param coverage_result a coverage object returned from \code{\link{package_coverage}}
+#' @param coverage_result a coverage object returned from
+#' 	\code{\link{package_coverage}}, or its data frame conversion
 #' @export
-zero_coverage <- function(coverage_result){
-  coverage_data <- as.data.frame(coverage_result)
+zero_coverage <- function(coverage_result) {
+  coverage_data <- if (is.data.frame(coverage_result)) {
```
jimhester: This test is unnecessary, if an object is already a `data.frame` `as.data.frame` will just return the object without doing anything if `x` is already a `data.frame` (Assuming the `row.names` are not `NULL`).

R/covr.R
kforner: I do not remember exactly, but it is caught by one of my new tests, whith a no-argument function. In fact formals(f) <- NULL crashes when the formals were already NULL.

R/summary_functions.R
kforner: ok, good to know

R/summary_functions.R
```diff
@@ -4,8 +4,9 @@
 #' @export
 percent_coverage <- function(coverage_result){
   coverage_data <- as.data.frame(coverage_result)
-  num_lines <- nrow(coverage_data)
-  num_not_zero <- sum(coverage_data$value != 0) # how many lines have a value that is not zero?
+  num_lines <- length(unique(coverage_data$first_line))
```
jimhester: Not sure about the motivation for these changes, they seem to make the intent less clear to me.

This function should probably just be made into two S3 functions
```r
percent_coverage <- function(coverage_result) {
  UseMethod("percent_coverage")
}

percent_coverage.coverage <- function(coverage_result) {
  sum(x > 0)  / length(x)
}

percent_coverage.data.frame <- function(coverage_result) {
  sum(coverage_result$value > 0) / NROW(coverage_result)
}
```

kforner: As I wrote in the introduction, the sum(coverage_result$value > 0) / NROW(coverage_result) is wrong because it assumes that all first_line are unique. I provided a test that shows it.

R/summary_functions.R
```diff
@@ -34,12 +40,11 @@ zero_coverage <- function(coverage_result){
 print.coverage <- function(x, ...) {
   df <- as.data.frame(x)
 
-  per_file_percents <-
```
jimhester: Now sure why this change was made.  With the S3 percent_defaults I mentioned above this can be simply
```r
per_file_percents <- tapply(df$value, df$filename, FUN = percent_coverage, simplify = TRUE)
```

docker_checker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+MAINTAINER karl.forner@quartzbio.com
+
+RUN Rscript -e 'install.packages("digest")'
+RUN Rscript -e 'install.packages("crayon")'
+RUN apt-get install -y libcurl4-openssl-dev sudo nano
+
+RUN Rscript -e 'install.packages("devtools")'
+RUN Rscript -e 'install.packages("roxygen2")'
+RUN Rscript -e 'install.packages("testthat")'
+RUN Rscript -e 'devtools::install_github("jimhester/rex")'
```
jimhester: `jimhester/rex` is not the main repository, this should be `kevinushey/rex`.

Also is there a rex feature needed in `covr` that is not available on the CRAN version of `rex`? (I do not recall one offhand)

If there is I should do a new `rex` release to CRAN.

docker_checker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM rocker/r-devel
+MAINTAINER karl.forner@quartzbio.com
+
+RUN Rscript -e 'install.packages("digest")'
```
jimhester: Is there a reason you can't install all the packages in one statement?  Also you should only need to install the packages in Imports 
```
    jsonlite,
    rex,
    devtools,
    httr,
    testthat,
    crayon
```

docker_checker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM rocker/r-devel
+MAINTAINER karl.forner@quartzbio.com
```
jimhester: Do you want to remain the maintainer for the docker stuff or should I be the maintainer?

tests/testthat/TestS4/test.Rproj
```diff
@@ -5,12 +5,13 @@ SaveWorkspace: No
 AlwaysSaveHistory: Default
```
jimhester: This file should be removed from git (not your fault though I am sure I inadvertently added it).

tests/testthat/test-trace_calls.R
```diff
@@ -5,10 +5,15 @@ test_that("one-line functions are traced correctly", {
   on.exit(options(keep.source = old))
 
   fun <- function(x) x + 1
-
   expect_equal(as.character(body(trace_calls(fun))[[2]][[1]]),
     c("::", "covr", "count"))
 
+  fun <- function() 1
+#browser()
```
jimhester: Please remove these debugging lines

docker_checker/Dockerfile
kforner: no, I just looked at your github and saw it. I'll try the CRAN version.

R/summary_functions.R
jimhester: I guess this depends on what you interpret percent_coverage to mean.  The
way I interpret it is percent of all traced expressions covered, so two
expressions on the same line are counted separately.  You seem to consider
percent_coverage as the percentage of lines covered, in which case your fix
works, but it was not my original intent.

This is ultimately my responsibility as I did not closely review the logic
for the pull request which introduced the percent_coverage function, I
should have made the distinction more clear.

docker_checker/Dockerfile
kforner: sure, it's the result of incremental changes. I'll clean it.

docker_checker/Dockerfile
kforner: no, you should be. Anyway this docker will never be published so we can put anything.

tests/testthat/TestS4/test.Rproj
kforner: yes I do not use rstudio ;)

R/summary_functions.R
kforner: Hmm I see. But the problem is that when you visually look at the results in the source files, if on the same line you have one expression that is covered and one that is not, how do you represent it ? 
In the same spirit, the developer would look at the missed lines, and it will be hard to grasp why a line is missed although he can be sure that the main expression on this line was executed, do you see what I mean ?

kforner: Anyway we could provide, via a parameter or different functions the two behaviors.

jimhester: I believe the https://github.com/MangoTheCat/testCoverage package
highlights only the portion of the line that includes the expression.  The
srcref contains the columns a given expression covers in the source file,
so it is possible to do.

coveralls only reports coverage per line, so it does make sense to have a
parameter, or different functions (line_coverage, expression_coverage
perhaps) to calculate each percentage.

kforner: so what do you want me to do ? restore your original functions and add new ones ?

jimhester: No thinking about it more lets add a parameter as you suggested and have
percent coverage by line be the default.

jimhester:
Looks good to me, thanks!

PR #31, opened 2015-01-14
BUG FIX: trace_calls() gives an error on a function with NULL as body.

HenrikBengtsson:
This pull request fixes the following problem:
```r
> foo <- function() NULL
> covr:::trace_calls(foo)
Error in if ((is.symbol(fun_body) || fun_body[[1]] != "{") && !is.null(attr(x,
:
  missing value where TRUE/FALSE needed
```


HenrikBengtsson:
Without fix: https://travis-ci.org/HenrikBengtsson/rmini/builds/46946487
With fix: https://travis-ci.org/HenrikBengtsson/rmini/builds/46947409

jimhester:
Could you add a test for this situation to [tests/testthat/test-trace_calls.R](https://github.com/jimhester/covr/blob/1c8f5d9548d53074b4ac7723f3cdabad5be5924c/tests/testthat/test-trace_calls.R)

You should be able to basically follow the previous examples, just test that the function is not changed from the original and doesn't error.

HenrikBengtsson:
Added test.

jimhester:
Thanks!

PR #47, opened 2015-01-29
Skip primitive functions as alias in package

renkun-ken:
The function check is incomplete. If an internal function of package is merely an alias of a primitive function, for example,

```r
set_names <- `names<-`
```

the function has `NULL` environment (e.g. `class<-` and `names<-`), the `replacement` function will end up with an error:

```
Error in as.function.default(c(value, if (is.null(bd) || is.list(bd)) list(bd) else bd),  : 
  use of NULL environment is defunct
```

Therefore, all primitive functions (or functions with `NULL` environment) should be skipped.

### DIFF
--- a/R/replace.R
+++ b/R/replace.R
@@ -1,7 +1,7 @@
 #' @useDynLib covr duplicate_
 replacement <- function(name, env = as.environment(-1)) {
   target_value <- get(name, envir = env)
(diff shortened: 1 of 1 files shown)


jimhester:
Seems reasonable for now, thanks!  We may want to track these better in the future, I will open an issue.

Issue #46, opened 2015-01-28
Files in subdirectories of src cause errors again


jonclayden:
After issue #33 was fixed, `covr` correctly worked with source files in subdirectories of `src`, but there seems to be another problem in the current version. After updating I see this when I try to check coverage in package `ore`:

```
> library(covr)
> package_coverage()  # from "ore" package root directory
Regular expression compilation : .....
Matching and related operations : .................
Compatibility with rex : .

Error in file(con, "r") : cannot open the connection
In addition: Warning message:
In file(con, "r") :
  cannot open file './src/iso8859_1.c': No such file or directory
```

The correct path is actually `./src/onig/enc/iso8859_1.c`. If it's relevant, `Makevars` says

```
OBJECTS_ENC = onig/enc/ascii.o onig/enc/iso8859_1.o onig/enc/unicode.o onig/enc/utf8.o

OBJECTS_ONIG = onig/regcomp.o onig/regenc.o onig/regerror.o onig/regexec.o onig/regext.o onig/reggnu.o onig/regparse.o onig/regposerr.o onig/regposix.o onig/regsyntax.o onig/regtrav.o onig/regversion.o onig/st.o

OBJECTS = main.o $(OBJECTS_ONIG) $(OBJECTS_ENC)

PKG_CPPFLAGS = -Ionig
```

jimhester:
Sorry we got a regression Jon, I will look into it and make sure to have a test for this situation going forward.

PR #61, opened 2015-02-28
Do not ignore leading path components when finding gcov files

jonclayden:
This fairly trivial change seems to fix issue #46.


jimhester:
That is good!  I would like to get a test case for a package with source files in sub directories as well.  If you have time to create one I would love to have it (you should be able to copy [tests/testthat/TestCompiled](https://github.com/jimhester/covr/tree/master/tests/testthat/TestCompiled) and modify it so we can ensure there are no regressions of this later on.

If you don't think you will have the time to add the test case let me know and I can handle it.

Thanks for taking the time to debug this and send the pull!

jonclayden:
I've added a test case. In the process I discovered that the fix wasn't actually so trivial, but I've amended it and all seems to be well now.

jonclayden:
Or perhaps not... I can't replicate the test failure on Travis locally, so I'm not sure what's going on. For the latest commit I guessed that its `gcov` was creating files in different places to mine, but that seems not to be the problem. The error, "attempt to set an attribute on NULL", suggests that the result from `parse_gcov()` is empty, but I don't know why that is. Sorry about this.

jimhester:
I will see if I can track it down tomorrow, thank you for the test case and
getting it this far, it is already a big help

jonclayden:
OK, I've managed to reproduce it within an Ubuntu 12.04 virtual machine. The version of `gcov` there behaves oddly. It ignores path elements in its arguments, so `gcov lib/simple.c` actually searches for `simple.gcno` within the working directory, not within `lib`. But, because the path is included in the `.gcno` file itself, running `gcov` from within `lib` also doesn't work. So the only solution that actually works is to copy the `.gcno` and `.gcda` into `/src` and then run `gcov` from there. I will try and put together a version of the code that does that, but by nature it's going to be a bit less clean. Of course, you may be able to find a tidier solution later.

jimhester:
There seem to be at least 3 different gcov versions with different behavior we need to support.

- gcov from earlier gcc versions
- recent gcc gcov
- clang gcov

frustratingly (as you have seen) they have different behavior with how they treat paths.

The `-p --preserve-paths` option seems to be what we want to do from https://gcc.gnu.org/onlinedocs/gcc/Invoking-Gcov.html.  But I am not positive it is available on all the above versions.

jonclayden:
Sure, I noticed that option, but still:

```
~/git/covr/tests/testthat/TestCompiledSubdir/src$ gcov -p lib/simple.c
simple.gcno: cannot open graph file
```

The `-p` option doesn't help here. The Travis build now seems to be failing due to a `devtools` version problem(?!), but on my test VM the package now passes its tests with my latest commit.

jonclayden:
Looks like `devtools` master is broken, so it's defaulting to installing from CRAN, which doesn't have a high enough version number.

jimhester:
Yeah I am going to remove the version requirement and get this merged.  No need to do anything!

jimhester:
Thanks for working on this Jon, hopefully this is the last time this issue rears it's ugly head.

jonclayden:
Glad to help!

PR #64, opened 2015-03-03
Added support for all Fortran source file extensions starting with f

helske:
I added support for other common Fortran file extensions (previously only `f` was recognized) by modifying `compiled` function so it captures all file extensions starting with `f` (just copied the `c` part) .  I also modified `package_coverage` function by adding `FCFLAGS' for Fortran 9x code. 

This seems to work with KFAS: https://coveralls.io/builds/2032310


jimhester:
Looks good, thanks!

PR #72, opened 2015-03-16
Add git info for unknown CI backends for coveralls

kirillseva:
This pr is a follow-up for https://github.com/jimhester/covr/pull/70
When using coveralls outside of supported CI backends, coveralls is not able to populate git info and keep track of branches, commit authors, etc. This PR adds git info for drone.io backed CI builds.

I'll write a blogpost later about how one can get blazing fast CI builds using drone and a wrapper around `devtools::test()` (builds that take less than a minute to run)


kirillseva:
Tested on my fork - works

kirillseva:
^ appveyor failed because it ran before you fixed it, I presume. Can you restart the build?

jimhester:
I think you would have to rebase/merge from master before it would build properly.  But don't worry about it I know why it failed, so no need to do so.

jimhester:
I think this idea has merit, however I have some comments.

1. While I love `magrittr`'s pipe operator for doing analysis, I don't think using it in package code is a great idea for the same reasons that using `subset`, `with` ect are bad ideas (non-standard evaluation). 
2. You may want to look at https://github.com/jimhester/covr/pull/66 which adds codecov support which already supports drone.io.  As it is getting largely the same information you are trying to get here it is probably better to re-use some code.  Although I do prefer your use of the switch statement with functions to retrieve the information rather than their if/else approach.
3. When parsing `git log` do it once and get all the information you need rather than running `git log` multiple times.
4. I would either gather all the information from the CI builds environment variables or none of it.  Also the environment variables you referenced don't seem to correspond to the ones the drone docs say they set http://docs.drone.io/env.html

kirillseva:
I'm actually using a self-deployed digitalocean droplet running drone, as opposed to drone.io, which somehow has different semantics... 
As such, I'm using as many CI env variables as I can. I can rewrite to use only `system('git ...')` commands, which should work with every CI platform.

As far as maggrittr goes, I think I makes code cleaner and more readable. Compare these:
```r
paste0("git log -n 1 --pretty=format:'", pattern, "'") %>%
system(., intern = TRUE) %>%
jsonlite::unbox(.)
```
vs
```r
jsonlite::unbox(system(paste0("git log -n 1 --pretty=format:'", pattern, "'"), intern = TRUE))
```
Which one would you prefer to read?

Same with git log. Calling it multiple times is not a performance bottleneck, but makes for cleaner code IMO. I'll refactor that in a new commit and see if new approach reads better.

jimhester:
I'd prefer the one that doesn't cause potential errors due to using non-standard evaluation.  I understand the benefits of the pipe operator for readability but it does not belong in package code.  

Calling git log repeatedly can be a performance bottleneck with large repositories and your code is not more readable than simply getting all the results.  Just use something like the following.

```r
  fmts <- c(
    id = "%H",
    author_name = "%an",
    author_email = "%ae",
    commiter_name = "%cn",
    commiter_email = "%ce"
  )

  structure(
    scan(
      sep="\n",
      what = "character",
      text=system(intern=TRUE,
        paste0('git log -n 1 --pretty=format:',
          paste(collapse="%n", fmts)
        )
      )
    ),
    names = names(fmts)
  )
```

If you are worried about delimiter conflicts with newline use one of the [ASCII delimiters](http://en.wikipedia.org/wiki/Delimiter#ASCII_delimited_text)

kirillseva:
thanks for comments! I'll refactor and push for another review

kirillseva:
Apparently my env vars are same as Jenkins uses. Apparently drone tries to be compatible

DESCRIPTION
```diff
@@ -5,10 +5,10 @@ Authors@R: "Jim Hester <james.f.hester@gmail.com> [aut, cre]"
 Description: Track test coverage for your package and (optionally)
     upload the results to coveralls.  Compatible with any testing
     methodology or framework.
-Depends: 
+Depends:
     R (>= 3.1.0),
     methods
-Imports:
```
kirillseva: trailing whitespace is superfluous :black_joker:

kirillseva:
What do I need to refactor to get this merged @jimhester ?

R/coveralls.R
```diff
@@ -57,9 +71,45 @@ to_coveralls <- function(x, service_job_id = Sys.getenv("TRAVIS_JOB_ID"),
...
+    commiter_email = "%ce",
+    message = "%s"
+  )
+  head <- lapply(structure(
+    scan(
+      sep="\030", # http://en.wikipedia.org/wiki/Delimiter#ASCII_delimited_text
+      what = "character",
+      text=system(intern=TRUE,
+        paste0("git log -n 1 --pretty=format:",
+          paste(collapse="%n", formats)
```
jimhester: if you are going to use ascii delimiters you need to also put them here

kirillseva:
```r
paste0("git log -n 1 --pretty=format:", paste(collapse="%n", formats), "\030")
```
?

R/coveralls.R
```diff
@@ -57,9 +71,45 @@ to_coveralls <- function(x, service_job_id = Sys.getenv("TRAVIS_JOB_ID"),
       "source_files" = res)
   }
 
+  if (!is.null(git_info)) payload <- c(payload, git = list(git_info))
```
jimhester: this can just be

```r
payload$git <- git_info
```

if git_info is `NULL` then it won't be added to the list

Also what happens when repo token is not null but git_info is?  Payload won't be declared on line 74

R/coveralls.R
jimhester: no the collapse argument.  If scan is expecting a separator of `\030` you need to have git log put a `\030` between each format as well i.e. `paste(collapse="\030"`

Right now you are outputting a newline separator from git log but trying to read using a ascii record separator.

R/coveralls.R
kirillseva: git_info will at least contain
```r
git_info$remotes$name == "origin"
```
But yeah, should probably error gracefully

jimhester:
This also needs some tests before it is merged, which could have caught your separator mismatch error above for instance.

R/coveralls.R
kirillseva: Github hid some diff lines. But payload will always exist with some defaults, and your coverage will be reported to coveralls (just without git info)
```r
payload <- if (is.null(repo_token)) {
    list(
      "service_job_id" = jsonlite::unbox(service_job_id),
      "service_name" = jsonlite::unbox(service_name),
      "source_files" = res)
  } else {
    list(
      "repo_token" = jsonlite::unbox(repo_token),
      "source_files" = res)
  }
```

jimhester: Ah understood, guess I should have expanded to the full file, sorry!

kirillseva:
Switched back to `'\n'` separator, as `git log ... %s` returns abbreviated commit message that fits in one string, and other fields should not have `'\n'`
Thanks for reviewing!

kirillseva:
Anything left unaddressed? I'd love to merge this and switch my builds back to `jimhester/covr`

jimhester:
I may tweak some things after merging, but it looks good.  Thanks for working through everything and for the contribution!

kirillseva:
Thanks for reviewing @jimhester !

PR #115, opened 2015-09-25
add 'no_try' argument

brodieG:
Adding a `no_try` argument so that `package_coverage` can run tests without wrapping them in a `try` call.

Also includes tests for this new feature.


brodieG:
Not sure what's going on with the failed tests.  Werker just seemed to hang in the dependencies step.  I had run `test_check` separately and all tests passed.

brodieG:
Actually, looks like the hanging was just client side.  The only issue I think I introduced was not updating the Rd file which caused a warning in R CMD check since there is now an "undocumented" argument (the roxygen markup was updated).  The rest of the errors seem to be R lint related in the master code.  Let me know if you're interested in this branch and if there are any other changes you would want me to make.  Otherwise I'll just let it be here.

jimhester:
Thanks Brodie!

The only major change I made to your code was renaming the argument to `use_try` rather than `no_try` as I find negative arguments harder to reason about.

PR #118, opened 2015-10-03
update 'use_try' logic to line up with argument intent

brodieG:
When changing from admittedly confusing 'no_try' to 'use_try', the
underlying logic needed to be negated as well to avoid a truly
confusing mess (which goes to highlight your point about avoiding
negative arguments to begin with...)


jimhester:
Thanks! silly mistake by me

PR #122, opened 2015-10-21
replaced the tryCatch() by a try() in try_unload because of potential…

kforner:
… problems with nested tryCatch() calls

cf http://stackoverflow.com/questions/33256075/odd-behaviour-of-nested-trycatch

When calling **package_coverage()** with a variant of mclapply which uses a tryCatch(), the **unloadNamespace(pkg)** fails (for good reason), but **try_unload()** fails to resume processing, so that the run_tests() function never returns the coverage results.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -315,7 +315,8 @@ run_tests <- function(pkg, tmp_lib, dots, type, quiet, use_try=TRUE) {
 }
 
 try_unload <- function(pkg) {
-  tryCatch(unloadNamespace(pkg), error = function(e) warning(e))
+  e <- try(unloadNamespace(pkg))
+  if (inherits(e, 'try-error')) warning(e)
 }
 
 process_examples <- function(pkg, lib = getwd(), quiet = TRUE) {


jimhester:
`try()` is implemented with `tryCatch()`, so that alone can't be the reason this isn't working as expected.
``` r
try
#> function (expr, silent = FALSE) 
#> {
#>     tryCatch(expr, error = function(e) {
#>         call <- conditionCall(e)
#>         if (!is.null(call)) {
#>             if (identical(call[[1L]], quote(doTryCatch))) 
#>                 call <- sys.call(-4L)
#>             dcall <- deparse(call)[1L]
#>             prefix <- paste("Error in", dcall, ": ")
#>             LONG <- 75L
#>             msg <- conditionMessage(e)
#>             sm <- strsplit(msg, "\n")[[1L]]
#>             w <- 14L + nchar(dcall, type = "w") + nchar(sm[1L], 
#>                 type = "w")
#>             if (is.na(w)) 
#>                 w <- 14L + nchar(dcall, type = "b") + nchar(sm[1L], 
#>                   type = "b")
#>             if (w > LONG) 
#>                 prefix <- paste0(prefix, "\n  ")
#>         }
#>         else prefix <- "Error : "
#>         msg <- paste0(prefix, conditionMessage(e), "\n")
#>         .Internal(seterrmessage(msg[1L]))
#>         if (!silent && identical(getOption("show.error.messages"), 
#>             TRUE)) {
#>             cat(msg, file = stderr())
#>             .Internal(printDeferredWarnings())
#>         }
#>         invisible(structure(msg, class = "try-error", condition = e))
#>     })
#> }
#> <bytecode: 0xa4a390>
#> <environment: namespace:base>
```
Turns out the problem isn't with the nested `tryCatch()`, it is with `warning(e)`. The latter is throwing an error (which is caught by the 'outer' `tryCatch()`). Changing your example accordingly gives us the expected result.
``` r
f <- function() {
  tryCatch(stop('Argh'), error = function(e) {
      message('f() error handler')
      warning(e$message)
    })
  message('computing')
  "Very Important Result"
}

tryCatch(x <- f(), error = function(e) str(e))
#> f() error handler
#> Warning in value[[3L]](cond): Argh
#> computing
x
#> [1] "Very Important Result"
```
Thank you for tracking this down, and for the simple example!

kforner:
ok, this renders my fix useless. Do you want me to pull another request ?

jimhester:
No I made the change bf9870a85446e28b07cb15761c37e4ad811fb490, thank you though!

kforner:
I saw that you already fixed the function, thanks !

PR #138, opened 2016-01-05
Exporting to_shiny_data()

JeanLescut:
The function to_shiny_data() can be very useful when one wants to create his own Shiny app


jimhester:
The issue with exporting this function is then I have to commit to keeping the API relatively constant. The data is also fairly specifically tailored to the shiny application within covr.

Could you explain your use case (possibly with an example application) to see if there is a more sutible method you could make use of? (`as.data.frame.coverage` or `per_line` come to mind).

JeanLescut:
My problem is very specific, so in fact you're right, I just use the ugly ":::".
Sorry about the merge request. And thank you again for the package !

PR #142, opened 2016-01-20
package_coverage() also returns test_resuts

JeanLescut:
I slightly modified covr.R to make package_coverage() also returns the results of the tests.
Also, I bypassed the source_dir() function in the covr package, to use the one in testthat... I shouldn't ?

I certainly dit things the very messy way, but it's just a starting point.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -224,6 +224,9 @@ package_coverage <- function(path = ".",
       coverage <- run_tests(pkg, tmp_lib, dots, type, quiet, use_try = use_try)
     }
   }
+  
+  test_results <- coverage[[1]] # Very dirty but you guys got the idea
+  coverage <- coverage[[2]]
 
   # set the display names for coverage
   for (i in seq_along(coverage)) {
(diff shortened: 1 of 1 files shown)


JeanLescut:
Please note that to retrieve the test results, one can either write 
`bquote(try(test_results <- devtools::test(pkg = .(testing_dir))))`
or :
`bquote(try(test_results <- testthat::source_dir(path = .(testing_dir))))`

Again, it is just an idea ...

jimhester:
covr doesn't assume any testing framework. It works with testthat, RUnit, testit or ad-hoc testing systems. The concept of test results don't really make sense for some of them.

This is not the first time this feature has been requested. The tests are run in an environment, so we could include that environment in our results, which would give users access to any objects created in that environment. I opened #144 to address this issue.

JeanLescut:
Thank you so much for your answer and your help for understanding what is going on. I will look forward the next version with great interest then.

realAkhmed:
I am facing the same issue. Need to examine the results of the tests before deciding whether to upload them or not.

@JeanLescut, thanks for the patch! Forking with your patch until #144

PR #146, opened 2016-01-28
Remove spurious text from README

gaborcsardi:
Or should it be a link with text "Coveralls"?


jimhester:
I think you are correct, not sure why that ended up there to be honest. Thanks!

PR #154, opened 2016-03-09
Impute srcrefs for subexpressions

### LINKED ISSUE #39: Conditional statements (if) reported as covered even if not run
Why is `do_something()` shown as covered in the following example:

```
if (FALSE)
  do_something();
```

Real-world example: For [the linked file](https://coveralls.io/files/436079701) there is only one test for which I'm pretty sure that `is.matrix(X)` is `TRUE` and `is.complex(X)` is `FALSE`. Also, the function doesn't end prematurely -- however, lines 33, 36 and 38 are indicated as covered.

krlmlr:
This allows collecting coverage for non-braced subexpressions in `if`, `for`, and `while` clauses. Works by imputing missing srcrefs on the fly (in `trace_calls()`) by analyzing `getParseData()` output.

Caveats:

- This doesn't work for me in `package_coverage()`, I don't know why. Perhaps `package_coverage()` somehow needs to know all srcrefs beforehand?
- Probably broken for non-ASCII characters.
- Percentage changes in one of the tests.

Fixes #39.

Test file:

```
f <- function() {
  if (FALSE)
    FALSE

  for (i in character())
    FALSE

  while (FALSE)
    FALSE

  repeat
    break
}

cv <- covr::function_coverage(f, f())
covr::shine(cv)
```

screenshot from 2016-03-09 16-32-53


jimhester:
This is **awesome** thank you for working on it, I had a couple aborted attempts and this has been a nagging problem I have been wanting to fix.

There shouldn't be any difference in srcrefs for `package_coverage()`.

However `package_coverage()` does use a subprocess, so you need to actually install the package with your changes before they will be used.

krlmlr:
package_coverage(): It seems that the package loading mechanism does something evil: It concatenates all R files to one, and adds `#line` directives. This beaks getParseData() on which my code relies. devtools::load_all() loads the code from the original files; would you mind using that instead of loadNamespace?

krlmlr:
Using load_all() breaks the S4 tests. Bummer...

jimhester:
I originally used `load_all()`, but as you saw S4 doesn't work correctly with it. I ran into the line directive issue for getting the source of each file as well. Here is the function which actually installs the code when you run `R CMD INSTALL` (https://github.com/wch/r-source/blob/b156e3a711967f58131e23c1b1dc1ea90e2f0c43/src/library/tools/R/admin.R#L205-L335).

To get around this I think we could just re-parse the text in the parent source reference to get the token data. Something like 
```r
txt <- as.character(parent_ref)
sf <- srcfile(txt)
parse(text = txt, srcfile = sf)
pd <- getParseData(sf)
```
Then you need to fix the line and column numbers from the tokens with those from where the if block starts.

krlmlr:
An alternative would be to fix getParseData() in this case. The "srcfile" attribute is an environment which contains an "original" object; I think the parse data belongs there, but is instead attached to the *last file* in the package. 

Test package: https://github.com/krlmlr/covr.dummy
Example output: http://rpubs.com/krlmlr/getParseData

Just posted to r-devel.

krlmlr:
Until this is fixed in R, I suggest we tweak the package loading process: Create a "last" file ourselves with an otherwise useless function, grab the parseData from there and put it where it belongs.

jimhester:
From your report it looks like the last srcref will always have the parse data attached. Could we just find that srcref and copy the parse data to the other srcrefs?

Figuring out the last file is slightly complicated due to Collate directives, locale issues etc.

krlmlr:
What if there are no objects in the last file? Think good ol' `zzz.R`.

krlmlr:
screenshot from 2016-03-10 16-38-48

jimhester:
Ok so we can tokens of the entire package including line directives from any srcref in the package with the following.

```r
get_tokens <- function(srcref) {
  getParseData(attr(parse(text = attr(getSrcref(srcref), "srcfile")$original$lines), "srcfile"))
}
```
Then you would need to fix the line numbers based on the line directives.

krlmlr:
Much easier, don't need to reparse at all. Working on a custom version of getParseData() to make it a tad faster. Functionality is in place already, tests pass, only lintr keeps complaining.

krlmlr:
Tests pass (AppVeyor requires an appveyor.yml file; if you delete the project there, it shouldn't show up here anymore). Character encoding still might be an issue (if we're ever looking at exact column position of the imputed srcrefs).

R/parse_data.R
```diff
@@ -0,0 +1,112 @@
...
+  switch(
+    as.character(x[[1L]]),
+    "if" = {
+      src_ref <- list(
+        NULL,
+        make_srcref(2, 4),
+        make_srcref(5),
+        make_srcref(6, 7)
+      )
+      src_ref[seq_along(x)]
```
jimhester: We need this to handle if's without else's right?  Might be worth adding a comment so it is clear why the `if` case is different than `for` and `while`.

gaborcsardi:
Wow. :+1:

R/parse_data.R
```diff
@@ -0,0 +1,112 @@
...
+}
+
+get_parse_data <- function(x) {
+  if (inherits(x, "srcref"))
+    get_parse_data(attr(x, "srcfile"))
+  else if (exists("original", x))
+    get_parse_data(x$original)
+  else if (exists("covr_parse_data", x))
+    x$covr_parse_data
+  else if (!is.null(data <- x[["parseData"]])) {
```
krlmlr: Like this:

```
  if (inherits(x, "srcref")) {
    get_parse_data(attr(x, "srcfile"))
  } else if (exists("original", x))
```

?

krlmlr:
Done. Should we wait for community feedback before merging? The changes here are rather invasive.

jimhester:
Well the current behavior is clearly wrong, so if we wait it shouldn't need to be long.

PR #162, opened 2016-03-19
Sort child parse data

### LINKED ISSUE #161: Srcref mismatch: Assignment in unbraced condition
```
f <- function(x) {
  if (x)
    a <- (1 < 2)
  else
    a <- (2 < 3)
}

covr::function_coverage(f, f(FALSE)) %>% covr::shine()
```

screencapture-127-0-0-1-7435-1458420350312

Doesn't happen with simple statements or if statements are braced. I'll take a look.

krlmlr:
in `impute_srcref()`, necessary for complex substatements. With tests.

Fixes #161.

### DIFF
--- a/R/parse_data.R
+++ b/R/parse_data.R
@@ -63,6 +63,7 @@ impute_srcref <- function(x, parent_ref) {
   stopifnot(length(pd_expr_idx) == 1L)
   expr_id <- pd$id[pd_expr_idx]
   pd_child <- pd[pd$parent == expr_id, ]
+  pd_child <- pd_child[order(pd_child$line1, pd_child$col1), ]
 
   line_offset <- parent_ref[[7L]] - parent_ref[[1L]]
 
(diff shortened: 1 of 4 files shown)


krlmlr:
Adapted failing test, coverage for TestSummary package:

screencapture-127-0-0-1-6088-1458462301957

I have a better test that uses zero_coverage() (86291aec5f8a2), but this requires adapting zero_coverage() first (b9eec4f6558e8). Should I include it here, or do you prefer a separate PR?

jimhester:
You can include them here that is fine.

I generally write tests on the result of `as.data.frame()` rather than using zero_coverage, but either works.

BTW I am currently working on the https://github.com/jimhester/covr/tree/v2 branch which uses a different method to load the tracing and retrieve the results which should be much more robust to test types and parallel code. I have merged both of your outstanding PR into it but it would be great if you could test it on a few packages to ensure it is working as intended. A benefit to the new approach is I believe the parse data is included appropriately so we don't need the fixing code.

krlmlr:
Done.

I believe you still want parts of get_parse_data() for better performance. Just look at how getParseData() is implemented. I've considered moving the sorting from this PR to get_parse_data(), but I think it's not worth the effort, because we only need a tiny bit of the data sorted.

It's great to see improvements in covr, I'll switch to v2 for my packages.

jimhester:
Thanks!

PR #168, opened 2016-04-21
Do not delete .def source files

jeroenooms:
Fixes https://github.com/jimhester/covr/issues/167.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -221,7 +221,7 @@ clean_objects <- function(path) {
   files <- list.files(file.path(path, "src"),
                       pattern = rex::rex(".",
                         or("o", "sl", "so", "dylib",
-                          "a", "dll", "def"), end),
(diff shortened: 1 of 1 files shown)


jimhester:
Yeah that seems correct, we should not be deleting user defined `.def` files and the automatically generated ones are already deleted by R.

FWIW the extensions used for this were originally from [`devtools::clean_dll()`](https://github.com/hadley/devtools/blob/b6a23be508043fd0ffc24316034b5e45bfc1a941/R/compile-dll.r#L54), probably `.def` should be removed there as well.

jimhester:
Thanks!

jeroenooms:
@jimhester yes it should be removed form devtools as well. Can you easily fix that or should i make a pr?

jimhester:
I would make a PR just in case there is a use case we haven't thought of so winston/hadley can see it.

PR #169, opened 2016-04-26
Condense display of code

kenahoo:
This change to the CSS makes the display of code much more compact.  I'm not much of a CSS developer though, so perhaps this isn't the best way to accomplish it.


jimhester:
I tweaked the CSS slightly, but the spirit of the PR remains.

Thanks, I should have done this long ago!

kenahoo:
Thanks!

PR #195, opened 2016-06-20
proof-of-concept (mc)parallel support.

kforner:
cf the last commit message.


jimhester:
Thanks for working on this Karl.

- I think this should functionality should just be used internally by `add_hooks()`, we can detect if a package depends on parallel with the following (assuming pkg is object returned by `as_package()`.
```r
uses_parallel <- any(grepl("\\bparallel\\b", pkg[c("depends", "imports", "suggests", "enhances", "linkingto")]))
```
- Maybe we should just add the save_trace call to the mc_exit body directly rather than using a closure, e.g.
``` r
lib <- "/tmp/lib"
b <- parallel:::mcexit
body(b) <- as.call(append(after = 1, 
  as.list(body(b)), bquote(covr:::save_trace(.(lib)), list(lib = lib))))
b
#> function (exit.code = 0L, send = NULL) 
#> {
#>     covr:::save_trace("/tmp/lib")
#>     if (!is.null(send)) 
#>         try(sendMaster(send), silent = TRUE)
#>     .Call(C_mc_exit, as.integer(exit.code))
#> }
#> <environment: namespace:parallel>
```

kforner:
> I think this should functionality should just be used internally by add_hooks(), we can detect if a package
> depends on parallel with the following (assuming pkg is an object returned by as_package()

I'm not a big fan when things are automagical and you have no control on them. Suppose for instance that you want to disable the fix on a package that use parallel. Or if you want to test the impact on some platform for a package that does not use parallel.

But you could mix the ideas, e.g.:
```
uses_parallel <- function(path) {
  pkg <- as_package(path)
  any(grepl("\\bparallel\\b", pkg[c("depends", "imports", "suggests", "enhances", "linkingto")]))
}

then
package_coverage <- function(path = ".", .............., fix_parallel_mcexit = uses_parallel(path), ......)
```

> Maybe we should just add the save_trace call to the mc_exit body directly rather than using a closure, e.g.

Yes indeed. Thanks , I did not know this trick !

jimhester:
I understand the need to disable this for testing, but as only a small fraction of packages are going to run into this issue I don't want to complicate the function signature.

We could have a global option to disable it, 
```r
if (isTRUE(getOption("covr.fix_parallel_mcexit", uses_parallel(pkg)))) {
  ...
}
```

kforner:
> We could have a global option to disable it, 

Yes, or/and a environment variable. which is convenient to modify the behaviour without changing the source code at all.

kforner:
So, what should I do ? patch via the body, and use the env variable ? Or do you prefer an option ?

jimhester:
I agree environment variables are nice, particularly on Travis. The main reason I would suggest an option is because covr already uses options for the exclusions. I suppose we could always support both as well...

kforner:
I do not understand why the appveyor build failed. Where can I find the error message ?

kforner:
I do not understand why the appveyor build failed. Where can I find the error message ?

kforner:
Any comment if you plan to merge the PR, and when ? 
I'd like to use it, and I'm reluctant on using forks of published packages.

R/covr.R
```diff
@@ -203,6 +203,9 @@ package_coverage <- function(path = ".",
   # add hooks to the package startup
   add_hooks(pkg$package, tmp_lib)
 
+  if (should_enable_parallel_mcexit_fix(pkg))
```
jimhester: I would prefer to put the conditional directly in `add_hooks()` rather than defining a new function to do this and replicating the read a write lines.

R/parallel.R
```diff
@@ -0,0 +1,50 @@
...
+uses_parallel <- function(pkg) {
+  any(grepl("\\bparallel\\b",
+      pkg[c("depends", "imports", "suggests", "enhances", "linkingto")]))
+}
+
+# consider in that order: the environment variable COVR.FIX_PARALLEL_MCEXIT,
+# the option covr.fix_parallel_mcexit, or auto-detection of the usage of
+# parallel by the package (cf uses_parallel()).
+should_enable_parallel_mcexit_fix <- function(pkg) {
+  var <- Sys.getenv('COVR.FIX_PARALLEL_MCEXIT')
```
jimhester: I would name the environment variable `COVR_FIX_PARALLEL_MCEXIT` and combine these all into one statement

```r
isTRUE(as.logical(
  Sys.getenv("COVR_FIX_PARALLEL_MCEXIT", 
    getOption("covr.fix_parallel_mcexit", 
      uses_parallel(pkg)))
```

jimhester:
Sorry Karl, I have been at conferences the past 10 days and haven't had a chance to look at it. If you could make the small tweaks mentioned and add a note to NEWS.md about the change I can merge this.

Thanks!

jimhester:
merged at starting at c23e6

Thanks for working on this Karl!

PR #197, opened 2016-06-24
Added initial cobertura support

wligtenberg:


R/cobertura.R
```diff
@@ -0,0 +1,181 @@
...
+          "branch-rate" = "0",
+          version = sessionInfo()$otherPkgs$covr$Version,
+          timestamp = as.character(Sys.time())),
+      parent = xmlDocument)
+  
+  # Add sources
+  sources <- XML::newXMLNode(
+      name = "sources", 
+      parent = top)
+  sourceFiles <- unique(covr:::display_name(cov))
```
jimhester: You can drop the triple colons here

R/cobertura.R
```diff
@@ -0,0 +1,181 @@
...
+        text = f,
+        parent = sources)
+  }
+  
+  # Add packages
+  packages <- XML::newXMLNode(
+      name = "packages", 
+      parent = top)
+  
+  packageName <- sub(".*?/(.+?)/R/.+?\\.[Rr]", "\\1", get("filename", attr(cov[[1]]$srcref, 'srcfile')))
```
jimhester: The coverage objects have a package object attached, so you can retrieve the package name with

```r
attr(x, "package")$package
```

jimhester:
Willhem thanks for working on this!

I would prefer to use the xml2 to generate the XML, but I can handle this part as the creation support in xml2 is new.

wligtenberg:
Ah, yes, I fear I am already becoming a bit old school. :)
I have created multiple XML generators in the past, before xml2 was there. Feel free to change library, this allowed me to get going quickly.

jimhester:
@wligtenberg I modified this to use xml2 in the wligtenberg-master branch, you can compare it to the common master at https://github.com/jimhester/covr/compare/master...jimhester:wligtenberg-master.

The main functional difference is xml2 does not currently support adding internal data type definitions, so I removed that from the output.

If you could test the example output and see if you get the expected results from cobertura it would be great!

PR #202, opened 2016-07-04
Removed useless and harmful on.exit(clear_counters()) call in package_coverage()

kforner:
cf https://github.com/jimhester/covr/issues/200


jimhester:
Thanks!

PR #204, opened 2016-07-06
Tweak logic for per_line() to handle closure definitions

### LINKED ISSUE #199: Coverage for "dynamically" Generated Functions
This one is pretty low priority since it is a fairly esoteric corner case:
```
make_fun_1 <- function() {
  function(x) {
    1 + 1
    2 + 2
    cat("fun1\n")
  }
}
fun1 <- make_fun_1()
```
With a test file that calls `fun1()`  does not show coverage of `1 + 1`, etc. even though that code is run.  I suspect this is different to issue #113 since `fun1` should be detected as a function in the namespace.  Likely the problem is that the function definition itself doesn't exist in a concrete manner that can be systematically defined in the source code.

I'm guessing the most likely outcome for this issue is "wontfix", but posting here just in case there is actually a simple resolution.

In case you're wondering about why anyone would ever want to do this: I happen to have a family of functions that are all almost the same and share the same formal arguments, so rather than defining each function separately I use a constructing function to avoid having to re-state a lengthy argument list every time (and risk messing it up for one of the functions when tweaking, etc.).

jimhester:
Fixes #199 

@brodieG This seems to be sufficient to fix the issue, taking advantage of the sorting when converting to a data frame and always taking the coverage from last expression.

screenshot 2016-07-06 10 03 40


brodieG:
Unfortunately I think this highlights lines as covered that shouldn't be (e.g. 35, 36, 41, 45) since they are only partially covered.  Would certainly be a _much_ easier solution.  Maybe there is a way to tweak, by tracking which elements are single line coverage and for those taking the min value encountered.  

Actually, maybe that just works.  Completely ignore any multi-line coverage info, and for each file-line, track min coverage for each single line statement.  One problem might be stuff like so:
```
1 +
1 +
2 * 4
```
Since this doesn't produce single line coverage.  That would break both this single line suggestion, and the byte coverage implementation I put in because of a simplification I made that assumed there would always be a single line core statement.  This can be worked around though.

I think the generalization is that we need to know when an expression is the innermost expression, which can be translated for our purposes into:

* single line expression, or
* multi-line expression with no sub-expressions within

These should be resolvable without resorting to byte coverage, though a bit more annoying.

jimhester:
Ok I tweaked the logic a bit, looks like it is now handling the corner cases you pointed out.

screenshot 2016-07-06 11 24 19

brodieG:
Yes, looks fantastic, thanks.  I think I misunderstood what the loop was doing.

brodieG:
One very low priority question: would it make sense for lines 2, 6, 9, 13 to show up as like 28 and 53, i.e. not relevant?  These are the `x <- function(){` bits.

jimhester:
I think it is better this way, it is possible you do want to know if those lines are uncovered. They can be covered by explicitly calling `make_fun_1()` in your tests.

brodieG:
Sounds good, thanks for addressing this so quickly.

PR #212, opened 2016-07-27
put back the tab function

maodunzhe:
Put the original tab api back to files.


jimhester:
Thanks! I did this in the same jquery chain (https://github.com/jimhester/covr/commit/c3c55dce954785cee3a40fbba941ed70fb8628f4) rather than a new query.

PR #220, opened 2016-09-03
close #219: incomplete gcov parsing

### LINKED ISSUE #219: Incomplete gcov parsing
Related to #164, but I think it's better to open a new issue, since it seems to be a parsing issue at the end. I run `package_coverage(clean = F)` and the `.gcov` output for [this file](https://coveralls.io/builds/7731688/source?filename=src%2Fentity.h) is the following:

```
        -:    0:Source:entity.h
        -:    0:Graph:/home/.../simmer/src/simmer_rcpp.gcno
        -:    0:Data:/home/.../simmer/src/simmer_rcpp.gcda
        -:    0:Runs:3
        -:    0:Programs:1
        -:    1:#ifndef ENTITY_H
        -:    2:#define ENTITY_H
        -:    3:
        -:    4:#include "simmer.h"
        -:    5:
        -:    6:#define SUCCESS 0
        -:    7:#define ENQUEUED -1
        -:    8:#define REJECTED -2
        -:    9:
        -:   10:// forward declarations
        -:   11:class Simulator;
        -:   12:
        -:   13:/** 
        -:   14: *  Base class. Every element in a simulation model is an entity.
        -:   15: */
    #####:   16:class Entity {
        -:   17:public:
    #####:   18:  CLONEABLE(Entity)
        -:   19:  
        -:   20:  Simulator* sim;
        -:   21:  std::string name;
        -:   22:  
    #####:   23:  Entity(Simulator* sim, std::string name, int mon)
    #####:   24:    : sim(sim), name(name), mon(std::abs(mon)) {}
    #####:   25:  virtual ~Entity() {}
    #####:   26:  virtual void reset() {};
    #####:   27:  int is_monitored() { return mon; }
        -:   28:  
        -:   29:private:
        -:   30:  int mon;
        -:   31:};
        -:   32:
        -:   33:#endif
```

Ok, no coverage, as shown in Coveralls. The issue comes out when you parse all these `gcov` files (`.gcov`, `.gcda`, `.gcno`) with `lcov`. The output of `lcov` is the following:

```
       1             : #ifndef ENTITY_H
       2             : #define ENTITY_H
       3             : 
       4             : #include "simmer.h"
       5             : 
       6             : #define SUCCESS 0
       7             : #define ENQUEUED -1
       8             : #define REJECTED -2
       9             : 
      10             : // forward declarations
      11             : class Simulator;
      12             : 
      13             : /** 
      14             :  *  Base class. Every element in a simulation model is an entity.
      15             :  */
      16          16 : class Entity {
      17             : public:
      18           0 :   CLONEABLE(Entity)
      19             :   
      20             :   Simulator* sim;
      21             :   std::string name;
      22             :   
      23        4126 :   Entity(Simulator* sim, std::string name, int mon)
      24        4126 :     : sim(sim), name(name), mon(std::abs(mon)) {}
      25        2861 :   virtual ~Entity() {}
      26           0 :   virtual void reset() {};
      27       11555 :   int is_monitored() { return mon; }
      28             :   
      29             : private:
      30             :   int mon;
      31             : };
      32             : 
      33             : #endif
```

Magic! :D From which I understand that `.gcov` files don't tell the whole story. Apparently, there is more info about those lines marked with `#####` in the `.gcda` and `.gcno` files (which are in binary format, BTW) that must be taken into account.

Enchufa2:
I'm exhausted, but this patch solves the issue. Proof: see my coverage [before](https://coveralls.io/builds/7713552) and [after](https://coveralls.io/builds/7732937) the patch. :)))

I'm going to bed...

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: covr
 Title: Test Coverage for Packages
-Version: 2.2.1
+Version: 2.2.1.9000
 Authors@R: c(person("Jim", "Hester", email = "james.f.hester@gmail.com", role = c("aut", "cre")),
     person("Willem", "Ligtenberg", email = "willem.ligtenberg@openanalytics.eu", role = "ctb"))
 Description: Track and report code coverage for your package and (optionally)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 2.2.1.9000 ##
+* Fix gcov execution: parse all files in the same subdir at the same time
+  instead of one by one (#220, @Enchufa2)
+
 ## 2.2.1 ##
 * Fix test broken with DT 0.2
 
--- a/R/compiled.R
+++ b/R/compiled.R
@@ -85,10 +85,15 @@ run_gcov <- function(path, quiet = TRUE,
   }
 
   gcov_inputs <- list.files(path, pattern = rex::rex(".gcno", end), recursive = TRUE, full.names = TRUE)
+  names(gcov_inputs) <- dirname(gcov_inputs)
+  gcov_inputs <- tapply(unlist(gcov_inputs, use.names=F), names(gcov_inputs), c)
   withr::with_dir(src_path, {
-    system_check(gcov_path,
-      args = c(gcov_args, gcov_inputs, "-o", dirname(gcov_inputs[[1]])),
-      quiet = quiet, echo = !quiet)
+    run_gcov <- function(src) {
+      system_check(gcov_path,
+        args = c(gcov_args, src, "-o", dirname(src[[1]])),
+        quiet = quiet, echo = !quiet)
+    }
+    lapply(gcov_inputs, run_gcov)
     gcov_outputs <- list.files(path, pattern = rex::rex(".gcov", end), recursive = TRUE, full.names = TRUE)
     structure(
       unlist(recursive = FALSE,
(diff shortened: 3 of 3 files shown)


Enchufa2:
Ready for your review.

jimhester:
@Enchufa2 Thank you for working on this, is could we get a minimal example of the problem so we can have a test for this?

Enchufa2:
Before this patch you were invoking gcov once per file:

```bash
gcov src/file1.gcno
gcov src/file2.gcno
...
```

and so on. But it seems that this way some interdependencies are missing in the final report. In fact, the manual says the following:

> If you invoke gcov with multiple input files, the contributions from each input file are summed. Typically you would invoke it with the same list of files as the final link of your executable.

That is,

```bash
gcov src/file1.gcno src/file2.gcno ...
```

and that's what this patch does. I agree that it would be better to have a test for this. I think that a method defined in one file and called in another one should do the trick: it should fail without this patch and success with it. But I can't work on this now. We can postpone this pull request, or merge it and open a new issue for the test. What do you think?

Enchufa2:
I saw that commit 586e65e already applies one gcov per directory, so this issue is fixed, right? There's no point in keeping this PR (as well as #226 and #219) open.

jimhester:
I think so, I believe I left them open because I wanted to verify the issues were fixed, if you can verify that was the case we can close them.

Enchufa2:
I switched to your covr repo in a new branch and [the coverage remains at 100%](https://codecov.io/gh/r-simmer/simmer/branch/covr), so it's definitely fixed.

jimhester:
Great thanks

PR #229, opened 2016-10-04
less code, more discussion

CSJCampbell:
Users may come to the vignette with no idea of what test coverage is. At least some description of the high level concept should be included in at least one vignette. Some of the examples seem to be extraneous detail of the development story, rather than examples of the current case. To make it as easy to understand as possible, try to avoid information that is not about the current behaviour of the tool.


jimhester:
Thanks for working on this Chris and for your patience in getting it merged, it reads much better now!

PR #237, opened 2016-12-06
Fixes for R_KEEP_PKG_SOURCE=yes.

kalibera:
It turns out that covr tests fail when packages are built with R_KEEP_PKG_SOURCE=yes, because the coverages obtained include also source line information from unexpected packages, such as from the R6 implementation. This patch fixes the covr tests to be robust against this, so that they work both with and without source references for packages.


jimhester:
I reproduced the problem by installing R6 with source references and we now filter only for results from the package being covered (824e2fc). Thanks again for the report and PR!

PR #247, opened 2017-01-25
icc code coverage support

chinqw:


R/covr.R
```diff
@@ -222,6 +222,25 @@ package_coverage <- function(path = ".",
 
   flags <- getOption("covr.flags")
 
+  ## BEGIN Oracle Contribution
+  # License: GPL-3 with additional permission for MIT under GPL-3 Section 7 as set forth in license documents.
+  #
+  # check for icc compiler
+  compiler <- get_compiler()
+  if (compiler == "icc")
```
jimhester: This should follow the conventions in the package and use [1TBS](https://en.wikipedia.org/wiki/Indent_style#1TBS)

R/covr.R
```diff
@@ -222,6 +222,25 @@ package_coverage <- function(path = ".",
...
+  {
+    if (length(getOption("covr.icov")) > 0L)
+    {
+      flags <- getOption("covr.icov_flags")
+      # clean up old icov files
+      unlink(file.path(pkg$path, "src","*.dyn"))
+      unlink(file.path(pkg$path, "src","pgopti.*"))
+    }
+    else
```
jimhester: Please but braces around all blocks, including single lines.

R/covr.R
```diff
@@ -426,3 +456,31 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
...
   writeLines(text = lines, con = load_script)
 }
+
+## BEGIN Oracle Contribution
+# License: GPL-3 with additional permission for MIT under GPL-3 Section 7 as set forth in license documents.
+#
+# check if gcc or icc is used
+get_compiler <- function()
+{
+  mkconf <- readLines(file.path(R.home(component="etc"), "Makeconf"))
```
jimhester: This will not take into account user or site wide settings in `Makevars`. I think the best way to do this is to call `R CMD CONFIG CC`.

R/covr.R
```diff
@@ -426,3 +456,31 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
...
+  compiler <- rex::re_matches(mkconf,
+                rex::rex(start, "CC = ",
+                         capture(name = "compiler", anything),
```
jimhester: Should probably just be `rex::rex(start, "CC", any_spaces, "=", any_spaces, capture(name = "compiler", non_spaces))`

R/covr.R
```diff
@@ -426,3 +456,31 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
...
+                         " ", anything))$compiler
+  compiler <- compiler[!is.na(compiler)]
```
jimhester: We should error if no compiler is found.

R/covr.R
```diff
@@ -426,3 +456,31 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
...
+
+  res <- try(system(paste(compiler, "--version"), intern=TRUE), silent=TRUE)
```
jimhester:
```r
tryCatch(
  system2(compiler, "--version"),
  error = function(e) {
    warning("cannot recognize a supported compiler")
})

File: R/covr.R [hunk 7]
```diff
@@ -426,3 +456,31 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
...
+  if (!inherits(res, "try-error"))
+  {
+    if (length(grep("gcc ", res)) > 0L)
+      return("gcc")
+    else if (length(grep("icc ", res)) > 0L)
+      return("icc")
+  }
```
jimhester: What about clang? Typically it is aliased as gcc, but it can also be explicitly specified.

R/compiled.R
```diff
@@ -99,3 +99,152 @@ run_gcov <- function(path, quiet = TRUE,
...
+  }
+
+  src_path <- normalize_path(file.path(path, "src"))
+  if (!file.exists(src_path)) {
+     return()
+  }
+
+  icov_inputs <- list.files(path, pattern = rex::rex(".dyn", end),
+                            recursive = TRUE, full.names = TRUE)
+
```
jimhester: If there are no icov inputs we can probably return early here.

R/compiled.R
```diff
@@ -99,3 +99,152 @@ run_gcov <- function(path, quiet = TRUE,
...
+  icov_profmerge <- getOption("covr.icov_prof", "")
```
jimhester: Return if this is not set?

chinqw:
Make changes according to the review comments and update DESCRIPTION and LICENSE files with new proposal. The new license proposal "License: MIT + file LICENSE + Portions GPLv3 with Additional Permission | GPLv3" in DESCRIPTION file seems to cause R CMD check a warning which is flagged by travis.

chinqw:
made changes according to review comments and updated license with a new proposal. It seems that the new license proposal 'License: MIT + file LICENSE + Portions GPLv3 with Additional Permission | GPLv3' causes R CMD check a warning which is flagged by travis.

jimhester:
I recently modified the covr license to be GPL-3 and you will need to sign the project CLA and remove the Oracle copyright clauses in order to contribute this code.

chinqw:
oracle license statement has been removed. Changes are sync'ed with master branch.

R/covr.R
```diff
@@ -293,10 +311,18 @@ package_coverage <- function(path = ".",
...
-    package = pkg,
-    relative = relative_path)
+  if (compiler == "gcc") {
+    coverage <- structure(c(coverage, run_gcov(pkg$path, quiet = quiet)),
+      class = "coverage",
+      package = pkg,
+      relative = relative_path)
+  } else {
+  # use icc compiler
+    coverage <- structure(c(coverage, run_icov(pkg$path, quiet = quiet)),
```
jimhester: Don't duplicate all of the code, just the object that differs.

```r
res <- if (uses_icc()) {
  run_icov(pkg$path, quiet = quiet)
} else {
  run_gcov(pkg$path, quiet = quiet) 
}
coverage <- structure(c(coverage, res), class = "coverage", package = pkg, relative = relative_path)
```

R/covr.R
```diff
@@ -429,3 +455,23 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
 
   writeLines(text = lines, con = load_script)
 }
+
+# check if gcc or icc is used
+get_compiler <- function() {
```
jimhester: I think this function should just check if `icc` is used and fallback to the current behavior otherwise. Also you cannot assume the `R CMD config CC` will have trailing spaces, using the  [word boundary](http://www.regular-expressions.info/wordboundaries.html) `\\b` anchor is a better solution.

```r
uses_icc <- function() {
  compiler <- tryCatch(
    paste(system(paste(R.home("bin"), "R --vanilla CMD config CC", sep="/"),
        intern = TRUE), collapse=""),
    error = function(e) "")
  grepl("\\bicc\\b", compiler)
}
```

chinqw:
suggested changes are done.

PR #249, opened 2017-01-26
Describe testthat integration

fangly:
as discussed at https://github.com/jimhester/covr/issues/248

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,29 @@ so that it is not revealed publicly.
 
 Also you will need to turn on coveralls for your project at <https://coveralls.io/repos>.
 
+## testthat ##
+If you would rather include unit tests for coverage in your [testthat](https://github.com/hadley/testthat) test suite, add a file `tests/testthat/test_coverage.R`:
+
+```
+context("coverage")
+
+test_that("Coverage", {
+  if (Sys.getenv("COVR_IS_RUNNING", FALSE)) {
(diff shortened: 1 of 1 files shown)


jimhester:
I think this would actually be best wrapped as a new expectation in `expectation.R`

```r
expect_coverage_gte <- function(percent = 90, ...) {
  if (isTRUE(Sys.getenv("COVR_IS_RUNNING", FALSE))) {
    testthat::skip("Coverage checks already running")
  }
  withr::with_envvar(
    c("COVR_IS_RUNNING" = TRUE),
    {
      cov <- package_coverage(...)
      avg <- percent_coverage(cov)
      testthat::expect_gte(avg, percent,
        label = round(avg, digits = 2),
        expected.label=paste0(percent, ", the required minimum coverage"))
    })
}

expect_coverage_gte(100, "digest")
#> Error: 96.36 is not more than 100, the required minimum coverage. Difference: -3.64
```

fangly:
Your idea is good, Jim, but this was merely an example. There are in fact two coverage tests that I am currently performing: requiring a minimum average coverage, and no uncovered functions. I am also generating a report. When using several coverage expectations, I think it makes sense to run the coverage calculation once and have it be used by the expectations (to avoid performing the coverage calculations multiple times).
Simply having a basic example somewhere in the documentation will be useful to anyone writing a coverage unit test.
Cheers,
Florent

PR #255, opened 2017-03-24
fix #254

### LINKED ISSUE #254: Testing, and function exclusions, fail with weirdly created functions
Here's a test case for a small package:

```
#' A foo
#'
#' With some foo
#'
#' @export
foo <- function(x, ...) UseMethod('foo')

make_foo_hux <- function() {
  assign('foo.huxtable', eval(bquote(function(x, ...) "hi")), envir = parent.frame() )
}

#' @rdname foo
#' @name foo.huxtable
#' @method foo huxtable
#' @export
NULL

make_foo_hux()
```

A complete source directory for the package is attached:

[packagefileprob.tar.gz](https://github.com/jimhester/covr/files/864373/packagefileprob.tar.gz)

When I run `package_coverage()` in the base directory, I get 

> Error in vapply(x, function(x) getSrcFilename(x$srcref, full.names = TRUE),  : 
>   values must be length 1,
>  but FUN(X[[2]]) result is length 0
> 

OK, fair enough, I'm doing something weird to create that function, it's not surprising that covr struggles with it. But when I run `package_coverage(function_exclusions = 'foo')` I get the same error. The traceback leads to `filter_non_package_files`.

hughjonesd:
This should be fine unless you really want to insist that all functions must have srcFilenames. An alternative would be unlist(lapply(...)).

### DIFF
--- a/R/display_name.R
+++ b/R/display_name.R
@@ -13,6 +13,6 @@ display_name <- function(x) {
 }
 
 filter_non_package_files <- function(x) {
-  filenames <- vapply(x, function(x) getSrcFilename(x$srcref, full.names = TRUE), character(1))
+  filenames <- vapply(x, function(x) if (length(fn <- getSrcFilename(x$srcref, full.names = TRUE)) > 0) fn else "", character(1))
   x[rex::re_matches(filenames, rex::rex(attr(x, "package")$path, "/"), "")]
 }


hughjonesd:
Failing check seems to be a weird issue with AppVeyor but not down to my code?

jimhester:
Sorry for the delay in getting this merged, thank you for the PR and your patience!

hughjonesd:
Hi Jim,

Thanks very much, but I think your latest commit (using vcapply)
reintroduces the problem!

Cheers,

David

hughjonesd:
Investigated a bit more. The issue is that the new function `get_source_filename` returns an empty string if it is passed in a length-0 vector. But, if a function was autocreated, the `srcref` attribute won't be zero length; it will be a `language` object like:
  `61L(5L, 61L, 39L, 5L, 39L, 796L, 796L)`

(Don't ask me why!)

The above returns `character(0)` from `utils::getSrcFilename`.

One fix would be to make `get_source_filename` check the length of `res`  rather than of `x`, but I am not sure whether that is OK with other contexts where you are using the new function.

hughjonesd:
Incidentally the test case in #254 is still a working test case for this.

jimhester:
Sorry I had a typo, the condition should be checking `res`, not `x`, should now be fixed.

hughjonesd:
Thanks once more.

David

jimhester:
Also added a test to prevent future regressions https://github.com/jimhester/covr/commit/79fe324db6ed111554f136f7e61539773a80342b

PR #271, opened 2017-07-15
preserve paths in gcov output

patperry:
I have a project with two C files having the same name, but in different directories. Without the "-p" flag, gcov will try to use the same file name for both `.gcov` files, clobbering one of the reports.

This patch adds the "-p" flag to gcov to preserve the file names in the reports. I've tested this on MacOS and Linux (Travis CI).

---

More info: the project is https://github.com/patperry/r-corpus
One instance of a name clash in that project are the files

```
src/termset.c
src/corpus/src/termset.c
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,10 @@
 * Improve behavior of `switch` coverage, it now supports default values and
   fall through properly.
 
+* Add `-p` flag to gcov command to preserve file paths. Fixes a bug where
+  gcov output didn't get reported when multiple compiled source files had
+  the same name (#271, @patperry)
+
 ## 3.0.0 ##
 * The covr license has been changed to GPL-3.
 * Set environment variable `R_COVR=true` when covr is running (#236, #268).
--- a/R/compiled.R
+++ b/R/compiled.R
@@ -64,7 +64,7 @@ run_gcov <- function(path, quiet = TRUE,
   withr::with_dir(src_path, {
     run_gcov <- function(src) {
       system_check(gcov_path,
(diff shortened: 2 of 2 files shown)


patperry:
A better solution perhaps to the one that I've proposed is to set the working directory to the source file directory before running gcov. I haven't tested that approach, though

jimhester:
We have gone back and forth on this a number of times, there are a number of different conflicting issues across gcc and clang gcov versions unfortunately.

patperry:
The issue is there for both clang (Mac OS) and gcc (Debian), and the "-p" flag fixes it on both platforms. Specifically:

Mac OS (clang)
```
Apple LLVM version 8.1.0 (clang-802.0.42)
  Optimized build.
  Default target: x86_64-apple-darwin16.6.0
  Host CPU: skylake-avx512
```
Debian (gcc):
```
gcov (Debian 6.3.0-16) 6.3.0 20170425
```
I don't know how other platforms behave.

patperry:
Any hope of getting this merged? If not, how should I proceed if I don't want to maintain my own fork of covr?

jimhester:
Could you rebase this on my current master and force push back to this branch, I can't really tell what the changes are with the current state.

```bash
git remote add upstream https://github.com/jimhester/covr.git
git fetch upstream
git rebase upstream/master
git push --force-with-lease
```

jimhester:
You will also need to add a note to NEWS.md with the change referencing this PR # and your GitHub username.

patperry:
Done, let me know if you need anything else.

jimhester:
Thanks!

PR #282, opened 2017-09-13
add function to convert coverage result to vector format

chen-liang-oracle:
Add a function trying to accept coverage result and output a vector format of the result for both file wise coverage and total package, please take a look


chen-liang-oracle:
@jimhester Hi Jim, I create a new function called coverage_to_list here to convert coverage result a list containing file coverage results as a vector and also added some comment. Please take a look and let me know your thought about that.

jimhester:
Thanks!

PR #299, opened 2018-02-14
Fix AppVeyor integration

RolandASc:
Tested with https://github.com/miraisolutions/winsendkeys.

See https://ci.appveyor.com/project/RolandASc/winsendkeys and possibly https://codecov.io/gh/miraisolutions/winsendkeys.

Added sink-file to capture return value and query elements.

### DIFF
--- a/R/codecov.R
+++ b/R/codecov.R
@@ -4,8 +4,7 @@
 #' `...`
 #' @param ... arguments passed to [package_coverage()]
 #' @param base_url Codecov url (change for Enterprise)
-#' @param quiet if `FALSE`, print the coverage and the query elements
-#' before submission.
+#' @param quiet if `FALSE`, print the coverage before submission.
 #' @param token a codecov upload token, if `NULL` the environment
 #' variable \sQuote{CODECOV_TOKEN} is used.
 #' @param commit explicitly set the commit this coverage result object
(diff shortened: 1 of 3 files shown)


RolandASc:
See also https://github.com/codecov/codecov-python/blob/master/codecov/__init__.py for reference. Note that I only cared about AppVeyor and did also not add the pull-request element to the query.

R/codecov.R
```diff
@@ -159,6 +159,10 @@ codecov <- function(...,
 
   coverage_json <- to_codecov(coverage)
 
+  if (!quiet) {
```
jimhester: I would prefer not to change this functionality.

README.md
```diff
@@ -73,6 +73,33 @@ after_success:
   - Rscript -e 'covr::codecov()'
 ```
 
+If you are using [Appveyor CI](http://ci.appveyor.com), and are not using
```
jimhester: This should just be something like 

```
on_success:
  - Rscript -e "covr::codecov()"
```
e.g. https://github.com/r-lib/fs/blob/3bb86eaf969b06f439970b982cdc849646332e46/appveyor.yml#L29-L30

R/codecov.R
RolandASc: ok, I have taken this back out

README.md
RolandASc: Simplified accordingly. Please go ahead if you would like to reword anything else in the ReadMe.

RolandASc:
Thanks for the comments - I agree about the default use case and have committed amendments.

jimhester:
Thanks for reworking things and for the original PR, looks good to me now!

PR #326, opened 2018-08-21
Add a separate trace_dir argument to add_hooks

siddharthab:
This allows us to run multiple tests in parallel with each test saving
the results in a different directory which can be configured through
an environment variable.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -375,7 +375,7 @@ package_coverage <- function(path = ".",
 
   coverage <- filter_non_package_files(coverage)
 
-  # Exclude both RcppExports to avoid reduntant coverage information
+  # Exclude both RcppExports to avoid redundant coverage information
   line_exclusions <- c("src/RcppExports.cpp", "R/RcppExports.R", line_exclusions, parse_covr_ignore())
 
   exclude(coverage,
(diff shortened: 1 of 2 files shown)


jimhester:
Are you missing the bit about the environment variable in the code? Or are you calling `covr:::add_hooks()` separately from `package_coverage()`?

R/covr.R
```diff
@@ -505,17 +505,25 @@ run_commands <- function(pkg, lib, commands) {
 # regardless of how the process terminates.
 # @param pkg_name name of the package to add hooks to
 # @param lib the library path to look in
+# @param trace_dir the directory where trace files will be saved; this can be
+#        an \code{expression} expression to evaluate the call at package load
+#        time, e.g. \code{expression(Sys.getenv("COVERAGE_DIR"))}.
 # @param fix_mcexit whether to add the fix for mcparallel:::mcexit
-add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE) {
+add_hooks <- function(pkg_name, lib, trace_dir, fix_mcexit = FALSE) {
```
jimhester: I would be ok with setting the default of trace_dir to be `trace_dir = Sys.getenv("COVERAGE_DIR", lib)`. Which would give you the behavior you were looking for just by setting the `COVERAGE_DIR` environment variable.

siddharthab: Done. Thank you.

jimhester:
Thanks!

Issue #327, opened 2018-08-24
Use GitLab internal pages


surmann:
Is it possible to use `covr` with the internal pages on GitLab? For me it would be fine, to have everything integrated. I think, the only missing feature, is the creation of local static html files.

[Here](https://about.gitlab.com/2016/11/03/publish-code-coverage-report-with-gitlab-pages/) GitLab describes how to publish an coverage report.

surmann:
Sorry, this is available via `report()`, I think.

jimhester:
From reading the document you should be able to use `report()` to generate the html which you can then serve.

If you get this working and would be willing to submit a PR with a `covr::gitlab()` function that makes it easy to do I think it would be useful.

surmann:
Thanks for your answer. I would like to submit something, to use gitlab with covr.

I worked on the yaml file for gitlab. [Here](https://gitlab.com/surmann/ODEnetwork) is an example. I just copied one of my projects and add a yaml with focus on code coverage. It works fine. More or less. Take a look at the [Result](https://surmann.gitlab.io/ODEnetwork/coverage.html). Unfortunately, the report messes up, and I have no glue why.

Do you have an idea?

If you like to test something I can give you write access to this repo.

surmann:
You can download the results from `covr` using artifacts within Gitlab. [Here](https://gitlab.com/surmann/ODEnetwork/pipelines) is the link to the pipeline. On the right is a button to download the artifacts. They show the same issue on the local machine. When I execute the `covr` code locally which I use in the [yaml](https://gitlab.com/surmann/ODEnetwork/blob/master/.gitlab-ci.yml) everything is fine.

What do you think?

surmann:
@jimhester 
I updated the installation of `covr` to the current github version. Unfortunately, the result looks still ugly. Do you have any hints how to work out the reason?

jimhester:
The css and javascript dependencies are not being loaded, it looks like because GitLab is serving them with the wrong mime types. e.g. this is what I get in the chrome dev console.

> Refused to apply style from 'https://surmann.gitlab.io/ODEnetwork/lib/datatables-css-0.0.0/datatables-crosstalk.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.

[...]

PR #330, opened 2018-08-29
add local report for gitlab

surmann:


surmann:
I have added some code with respect to https://github.com/r-lib/covr/issues/327#issuecomment-415741719

surmann:
@jimhester Is the AppVeyor issue a problem of my commit? It appears that devtools has a conflict with the R version.

jimhester:
could you try rebasing your commit against the current master? I updated the appveyor script, which I think should fix the problem. https://github.com/r-lib/covr/commit/1b349951c524e393147841ca15de3f8cdcc3057d

PR #342, opened 2018-11-06
Support to execute code coverage with no batches

javierluraschi:
Troubleshooting travis can be hard if not data is being printed, this is the behavior for `R CMD BATCH` but at times, one wants to print the output to the console of the tests that execute. This PR adds support for `batch = FALSE` to execute with plain `R -e 'command'` to be able to print output to console as code coverage executes.

### DIFF
--- a/R/codecov.R
+++ b/R/codecov.R
@@ -5,8 +5,8 @@
 #' @param ... arguments passed to [package_coverage()]
 #' @param base_url Codecov url (change for Enterprise)
 #' @param quiet if `FALSE`, print the coverage before submission.
-#' @param batch whether to execute commands using \code{R CMD BATCH}, defaults to
-#' \code{TRUE}.
+#' @param batch whether to execute commands using \code{R CMD BATCH} or
+#' \code{R -e '<command>'}, defaults to \code{TRUE}.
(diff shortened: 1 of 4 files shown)


jimhester:
Are you running extra commands on during the coverage apart from your standard tests? That is the only time this `R CMD BATCH` code is used.

The tests are not run by `R CMD BATCH`, they are run by `tools::testInstalledPackages()`, so this change will not affect the behavior. e.g. see

https://github.com/r-lib/covr/blob/278bc8b6b41e00f36529a685dff2850a8b0f3710/R/covr.R#L348-L353

Unfortunately I really want to continue using `tools::testInstalledPackges()`, as it mimics most closely the behavior of how tests are run during `R CMD check`.

javierluraschi:
@jimhester Yes, I'm running extra commands with something like `code = 'library(sparklyr); source("testthat.R")'`, so mostly not using `R CMD build` at all while running tests.

I find `R CMD build` really hard to debug so I'm mostly running `testthat` by myself in travis which is working quite well but, need this PR (or a similar mechanism) to let the output flow to the console while the tests execute.

jimhester:
Going to close this for now, we can re-open if needed.

PR #347, opened 2018-12-08
Closes #340; quick clarification of #nocov in src

### LINKED ISSUE #340: Proper nocov on src code
Discussion recently at `data.table`

https://github.com/Rdatatable/data.table/pull/3108

It seems one can "spoof" `nocov` into working on `src` code (in our case, C) by using

```
// # nocov
// # nocov start
// # nocov end
```

in compiled code.

Is this the intended API for doing so?

We noted there that `covr.exclude_pattern` & friends allow an entryway into a more natural version:

```
// nocov
// nocov start
// nocov end
```

It seems like it'd be more natural to allow one `pattern` for R code and another (or others) for code in other languages...

More generally it'd be helpful to see some comment about this in the documentation or `README`; I'll admit I haven't read every line of `man/` but I checked some likely places (`README`, `?exclusions`, `?file_coverage`, ...) and came up blank.

MichaelChirico:


jimhester:
Thanks!

Issue #375, opened 2019-06-20
how to skip clean_objects in covr::package_coverage ?


jpritikin:
You have

```
    clean_objects(pkg$path)
    withr::with_makevars(flags, assignment = "+=", utils::install.packages(repos = NULL, 
        lib = tmp_lib, pkg$path, type = "source", INSTALL_opts = c("--example", 
            "--install-tests", "--with-keep.source", "--with-keep.parse.data", 
            "--no-multiarch"), quiet = quiet))
```

If I already compiled the objects with the appropriate --coverage flags then I don't want to recompile them. However, I don't see how to avoid `clean_objects`. Please advise.

jimhester:
Fixed by https://github.com/r-lib/covr/pull/376

PR #376, opened 2019-06-20
Add option to skip clean_objects

jpritikin:
Here's how I solved #375. You probably have a better solution, but this is tested and works.


jimhester:
Thanks, I think this is fine, but can we name the argument `pre_clean` and make the default `TRUE`? Also can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jpritikin:
How's it look now?

jpritikin:
How to get this merged? These test failures don't look like my fault.

jimhester:

Thanks!! You are da bomb!

PR #393, opened 2019-08-20
Add pkgdown link to DESCRIPTION URL

batpigandme:
This will allow links to covr functions from other packages to go to the covr docs, rather than rdrr.io or rdocumentation.org.


jimhester:

Thanks!! You are da bomb!

PR #396, opened 2019-09-20
typo in manual

MichaelChirico:
I think this is what was intended...


jimhester:

Thanks!! You are da bomb!

Issue #413, opened 2019-12-14
Output to SonarQube Generic Data


PedroGuerraPT:
Hi, I'm trying to load covr stats into SonarQube using the `sonar.coverageReportPaths` parameter, along with the following instructions to generate the report:

```
cov_pkg_1 <- covr::package_coverage("./Package1")
cov_pkg_2 <- covr::package_coverage("./Package2")

covr::to_cobertura(cov_pkg_1, "coverage_pkg_1.xml")
covr::to_cobertura(cov_pkg_2, "coverage_pkg_2.xml")
```

Unfortunately it seems that the file format is not suitable for SonarQube. Would it be possible to implement a function that outputs a file according to the [supported format](https://docs.sonarqube.org/latest/analysis/generic-test/)?

Best regards,
Pedro

jimhester:
We would be happy to review a pull request to implement this feature, but we have no plans to work on it ourselves.

PedroGuerraPT:
Thanks for the reply Jim, we'll try to give it a shot and see if we can come up with a good implementation.

Delfic:
@jimhester Hey Jim, thank you for your time. We have added a simple function to extract the report with the format we need. We also included a unit test so that the function is also covered.
When you have the time we would love a review :)

PedroGuerraPT:
Implemented with #414 and #416

vladimirobucina:
@PedroGuerraPT  How do you implement this for example in a CI step?? I'm trying to run this, by setting `sonar.coverageReportPaths` but it's not working. sonarqube is just analyzing that file as if that's the file it should be looking at. It's not considering it as a report file.

PedroGuerraPT:
@vladimirobucina Unfortunately it has been quite a while since I last implemented this in Sonarqube with CI/CD, so I do not recall the implementation details anymore... 😕

PR #416, opened 2019-12-20
Added documentation to to_sonarqube.

Delfic:
While the solution to #413 worked, I forgot to run devtools::document().
Here is the updated documentation.


jimhester:
Thanks!

PR #455, opened 2020-11-22
Switch github actions to Ubuntu 20.04

billdenney:
I was looking for how to run covr with github actions, and I noticed a warning in your action log that github will soon switch to Ubuntu 20.04.  Here is a PR to make that switch.

(Thanks for covr!)

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -21,11 +21,11 @@ jobs:
           - {os: windows-latest, r: '3.6'}
           - {os: macOS-latest, r: '3.6'}
           - {os: macOS-latest, r: 'devel'}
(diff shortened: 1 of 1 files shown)


billdenney:
That wasn't as clean as I hoped it would be.

There appears to be an issue where the libicui18n.so.55 library for `stringi` isn't there.  I think that the error stems from the fact that there is now a libicui18n.so in libicu-dev package on the Ubuntu side.  I don't know how to fix that installation for the GitHub action.

jimhester:
It is because you also would need to switch the relevant RSPM locations. 

The warning is just informational for builds using `ubuntu-latest`, so there is nothing that needs to be changed.

Thank you however for taking the initiative and submitting the pull request, I really appreciate it!

billdenney:
I'm glad that it's not an issue.

Issue #459, opened 2020-12-12
`parse_gcov` assumes that produces source file names are escaped regular expressions


hsloot:
The function `parse_gcov` tries to skip all files for which `source_file` is not below root `package_path`. It does that by using a regular expression. 

https://github.com/r-lib/covr/blob/ebf0e7fc215ef91aeeb43ffcbe945e116e6e0187/R/compiled.R#L14

However, `rex::regex` assumes that the given expression is already an escaped regular expression. That might not be always the case - MacOS is allowing almost all Unicode characters in filenames. In my case, a parent folder contains a "+" (that, unfortunately, cannot be changed). 

It should be able to solve this by replacing the relevant expression with 

```r
grepl(rex::rex(start, rex::regex(rex::escape(paste0(package_path, collapse = "|")))), source_file)
```

Possible link to #448.

jimhester:
Yes, your fix looks correct, would you be interested in submitting a pull request to fix this?

PR #460, opened 2020-12-14
Fix regex bug in `parse_gcov`

hsloot:
Proposed fix for #459. Regular expression special characters in package directory will now be escaped. I added a small test which replicated the originally observed bug.


jimhester:
Looks great, thanks!

Thank you!

Issue #462, opened 2021-01-19
Feature Request: Recording srcrefs of test expressions & link to covr traces


dgkf:
It would be super helpful if, in addition to the srcrefs of the code that gets executed by each test, we could also record the srcrefs of the unit tests that trigger their evaluation.

# Motivation

I'm helping to develop an R package validation strategy, and one feature that would be tremendously valuable would be the ability to link individual unit tests to the traces that execute during the test execution. By linking the annotated test behavior to the trace that is executed, we can automate the construction of a traceability matrix, an expected part of any validation report. 

# Work so far

I've built up a proof of concept in a fork at [dgkf/covr@dev/testrefs](https://github.com/dgkf/covr/tree/dev/testrefs). However, to keep changes minimal, I've recorded a list of srcrefs (called `testrefs`) associated with the test for each and every trace (so that the length of `$testrefs` will be equal to `$value`).

```r
covr_result <- covr::package_coverage("./example")
covr_result[[1]]
# $value 
# [1] 1
#
# $srcref
# return("example result")
#
# $functions
# [1] "example"
#
# $testrefs
# $testrefs[[1]]
# test_that("example function communicates wip structure.", {
#   expect_silent(example())
# })
```

In my benchmarks, I found that execution time is minimally affected (~5% overhead). However, one big limitation of this approach is how much redundant information is recorded. Each and every call in the stack is recording a copy of the test trace. This results in quite large coverage objects (~50-100x tmp file size, and coverage object.size). There are certainly ways of collecting this data more thoughtfully, but it would affect more of the covr codebase.

# Proposed Changes to `coverage` Data Structure

If there's interest in including this additional data, I would propose a data structure that

1. Keeps existing data in the same place to minimize impact to existing `covr` code
2. Adds a list of test srcrefs as an attribute to the `coverage` object (`attr(,"tests")`)
3. Add a `tests` matrix  to each item within the `coverage` object list, which has a row per execution of the trace and columns for test number (corresponding to index in the test srcrefs attributes) and stack depth into the target package during execution (allowing differentiation of calls directly from the test from those that are called from within the package). 

```r
> unclass(covr_result)
# $example.R:1:2:3:4:5:6:7:8
# $`example.R:1:2:3:4:5:6:7:8`$value
# [1]  1
# $`example.R:1:2:3:4:5:6:7:8`$srcref
# return("example result")
#
# $`example.R:1:2:3:4:5:6:7:8`$functions
# [1] "example"
#
[...]

PR #463, opened 2021-01-28
Feature Addition (Addresses #462): Record test srcrefs and link to covr traces

dgkf:
# Overview

As described in #462, this PR adds functionality to capture data about the executing test while capturing coverage traces. This adds two pieces of additional data to the coverage object:

### `attr(,"tests")`

During collection, this is captured under `.counters$tests` (the `.counters` environment namespace uses srcref "keys" as names, so the `test` name can be confidently used without risk of conflicts.) This object is lifted out into an attribute in `package_coverage`. It includes curated calls from the call stack during the execution of a test. Calls are included if their source is within the tests directory or is the last call prior to entering the package namespace (before the first `count` call).  

```r
options(covr.record_tests = TRUE)
cov <- covr::package_coverage(testthat::test_path("testFunctional"))
attr(cov, "tests")
# $`./test-a.R:2:3:2:25:3:25:2:2`
# $`./test-a.R:2:3:2:25:3:25:2:2`[[1]]
# test_that("regular function `a` works as expected", {
#   expect_equal(a(), TRUE)
#   expect_equal(b(), FALSE)
# })
# 
# $`./test-a.R:2:3:2:25:3:25:2:2`[[2]]
# expect_equal(a(), TRUE)
# 
# $`./test-a.R:2:3:2:25:3:25:2:2`[[3]]
# a()
# 
# 
# $`./test-a.R:3:3:3:26:3:26:3:3`
# $`./test-a.R:3:3:3:26:3:26:3:3`[[1]]
# test_that("regular function `a` works as expected", {
#   expect_equal(a(), TRUE)
#   expect_equal(b(), FALSE)
# })
# 
# $`./test-a.R:3:3:3:26:3:26:3:3`[[2]]
# expect_equal(b(), FALSE)
# 
# $`./test-a.R:3:3:3:26:3:26:3:3`[[3]]
# b()
```

### `<coverage object>$<srcref key>$tests`

In addition, each covr trace keeps a listing of which tests result in the trace execution including the index of the corresponding tests in `attr(cov, "tests")` and the stack depth into the package at which the trace is executed:

```r
cov$`a.R:6:7:6:18:7:18:8:8`$tests
#      test depth
# [1,]    1     1
```

## Change log

- Added [`covr.record_tests` option](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L69-L71), which must be `TRUE` before additional information is used
- Added [call to `count_test`](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/trace_calls.R#L108-L136) from `count`, if option "covr.record_tests" is `TRUE`
- Added [`.current_test` environment object](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/trace_calls.R#L106) to the `covr` namespace (like `.counters`), which is used for capturing the current test during each trace count. It is invalidated if the executing call srcref changes and a new test trace, test frame and srcref take its place as the current test.
- Introduced an internal [`as_coverage`](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L109-L127) function, in order to centralize handling of the .counters$test value.
- [Made `merge_coverage` an S3 generic function](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L529-L534), dispatching on a character vector of filepaths or a list of coverage objects. This just makes it [easier to test](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/tests/testthat/test-record_tests.R#L121) without having to write coverage objects out to files. 
- [Added handling for `<coverage obj>$tests` consolidation in `merge_coverage`](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L552-L565)
- [Added new insertion into package startup](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L653), propagating `options(covr.record_tests = .(getOption("covr.record_tests")))` through to package startup within a test installation.
- [Added itemization of `covr` options](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L19-L74) to `?covr` package documentation. This isn't directly related to test srcrefs, but seemed like a gap I could address in the process. I tried my best to document the options whose behavior, pulling comments where they exist already somewhere in the code base when possible or when I could confidently infer from the source code. There were quite a few (mostly pertaining to gcc and icc) that I left blank. 
- Updated NEWS
- Bumped version number (roxygen version also got bumped to 7.1.1 while I was rebuilding the package, can revert if needed, but I'd need to downgrade my roxygen install to rebuild docs). 

# Design Considerations 

Just a couple design decisions that I want to raise to your attention for review:
1. I use `$tests` within `.counters`. I think this is safe, knowing that srcref keys will always take the form `a:1:2:3:4:5:6:7:8`, but it's always a bit disconcerting to attribute so much meaning to value names. 
2. Is there perhaps [a better way to propagate options to the child environment](https://github.com/dgkf/covr/blob/c32d9aee3c43794686e375a9d21478d596d814a6/R/covr.R#L652-L656) in which tests are executed?

# Benchmarks

I did a bit of benchmarking. Across the board, execution time was about 5-10% slower with `options(covr.record_tests)` enabled, and object size (using `lobstr::obj_size`) was ~5x the size. 

# Future Work

- There's room for improvement in how it handles packages tested with RUnit (more work needed to effectively build a test trace)
- Continue to test on a wider sampling of packages


jimhester:
It might be worth looking at how this handles the tests in data.table and knitr. Both of them do not use testthat or RUnit.

dgkf:
Thanks @jimhester! I've been trying to dig up good example packages and will definitely try these out. 

Regarding the CI checks, I'm able to successfully run tests on my machine (macOS R 4.0.3). I tried testing on an older R install (3.6.2) as well but ran into curl dynamic library issues - I think these are just issues with my machine, but can explore further if that might be helpful. At least for the linux builds, the issue seems to be with dependency installs. 

Do you have any suggested actions I should take to get builds passing?

dgkf:
I think I've added some reasonably generalized fallbacks for trying to collect the test traces. In trying to test an RUnit package ([`yaml`](https://github.com/viking/r-yaml)) and a bespoke testing package ([data.table](https://github.com/Rdatatable/data.table)), I encountered a couple other ways that srcrefs might be captured.

Just to document the process a bit:

### `testthat`

Packages tested with `testthat` are evaluated with the `./tests` directory as the working directory. If any srcrefs are from a relative filepath, we assume we're using testthat and any relative srcref filepaths can be assumed to be srcrefs of tests.

#### Example captured trace (using package, [praise](https://github.com/rladies/praise))

```
> cov <- package_coverage("praise")
> attr(cov, "tests")[1]
$`./test.R:17:3:17:32:3:32:17:17`
$`./test.R:17:3:17:32:3:32:17:17`[[1]]
test_that("template without praise word", {
  str <- "This is just a random string"
  expect_equal(praise(str), str)
  expect_equal(praise(""), "")
})

$`./test.R:17:3:17:32:3:32:17:17`[[2]]
expect_equal(praise(str), str)

$`./test.R:17:3:17:32:3:32:17:17`[[3]]
praise(str)
```

### `RUnit`

If no srcrefs in the call stack are relative paths, we check to see if any of the srcrefs that are within the temporary directory that covr installs the package into (`.libPaths()[[1]]` during coverage execution). 

#### Example captured trace (using package, [yaml](https://github.com/viking/r-yaml))

```
> cov <- package_coverage("r-yaml")
> attr(cov, "tests")[1]
$`/private/var/folders/123abc/T/Rtmp123/R_LIBS123/yaml/tests/test_as_yaml.R:121:3:121:91:3:91:121:121`
$`/private/var/folders/123abc/T/Rtmp123/R_LIBS123/yaml/tests/test_as_yaml.R:121:3:121:91:3:91:121:121`[[1]]
result <- as.yaml(list(foo=list(bar=list('foo', 'bar'))), indent.mapping.sequence = TRUE)
```

### Other

Finally, if all else fails, just save the whole call stack up until the call to `covr:::count`.

#### Example captured trace (using package, [data.table](https://github.com/Rdatatable/data.table))

```
> cov <- package_coverage("data.table")
> attr(cov, "tests")[2]
[[1]]
[[1]][[1]]
data.table(a = 1:2)
```

> For testing `data.table`, I removed all but one test script. Otherwise it takes quite a while to run. 

--- 

I think that these address a reasonable span of cases, but if there are others that I should look into I'd be happy to address them. 

@jimhester, I see that CLAassistant picked up a couple different emails in my commits. I would prefer to have all the commits under one email. Would you mind if I reintroduced a new PR after squashing the changes and re-committing under only one email?

jimhester:
You shouldn't need to open a new PR, you should be able to change the email on the commits with [interactive rebase](https://stackoverflow.com/a/1320317/2055486).

Alternatively if you just add the other email to your list of emails on GitHub that would also solve the issue I think.

jimhester:
> Packages tested with testthat are evaluated with the ./tests

This is not quite correct. `R CMD check` (and covr) executes all packages tests with the working directory in the tests directory. testthat further changes the working directory to `tests/testthat` when running its tests.

dgkf:
Thanks @jimhester - will update this in the description.

dgkf:
@jimhester - I went ahead and just removed the working directory behavior. Now, both `testthat` and `RUnit` traces are picked up by filtering for calls whose source files are within the temporary library (the second method [documented above](https://github.com/r-lib/covr/pull/463#issuecomment-775406304)). The relative directory as a way of subsetting the call stack felt like a hack at the start, so I'm happy to replace it with something more concrete. 

I edited my previous commits to all be from the same email, and the CLA looks good now. (Just a note, the CLA mentions the project is "current located at https://github.com/jimhester/covr or its successors or assigns" - wanted to raise that to your attention in case it's an oversight in the r-lib migration that needs to be updated).

Otherwise, I think this is in a good place unless you have other behavioral concerns. I looked into the failing CI builds, but I think these failures are holdovers from current master [db5423f](https://github.com/r-lib/covr/commit/db5423f8e549b5ecb60cb2f2301d74c536b7062e).

dgkf:
As far as I can tell, the builds pass with the exception of an issue that also currently exists in master - a dependency on `digest` that is unavailable in 3.2. 

@jimhester - is there anything more that you'd like me to look into before considering this PR?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # covr (development version)
 
+* Added `covr.record_tests` option. When `TRUE`, this enables the recording of
+  the trace of the tests being executed and adds an itemization of which tests
+  result in the exeuction of each trace. For more details see
+  `?covr.record_tests`
```
jimhester, line 6: Can you add `(@yourname, #issuenumber)` to this news bullet? And also add yourself to the `Authors@R` section of the package (just your name, no need for an email)

dgkf: Thanks - Added in 92d944d83e7b91f5044ceb779b4f1b8b3a4b7ea9, d96227e4d55c063b7fc13d874f855f563e586ad8

jimhester:
Thanks for working on this, I think it looks good as is.

krlmlr:
This seems to break tests: `git bisect` identified this pull request as a source of test failures that occur in a very similar way on the main branch. Checks seem to have passed for this PR, was this a change in the testing environment (base R, testthat, ...)?

krlmlr:
I'm seeing:

```
Failure (test-record_tests.R:13:3): covr.record_tests causes test traces to be recorded
length(attr(cov_func, "tests")) is not strictly more than 0L. Difference: 0

Failure (test-record_tests.R:14:3): covr.record_tests causes test traces to be recorded
length(attr(cov_func, "tests")[[1]]) is not strictly more than 0L. Difference: 0

Failure (test-record_tests.R:19:3): covr.record_tests records test indices and depth for each trace
ncol(cov_func[[1]]$tests) not equal to 2L.
target is NULL, current is numeric

Failure (test-record_tests.R:20:3): covr.record_tests records test indices and depth for each trace
colnames(cov_func[[1]]$tests) not equal to c("test", "depth").
target is NULL, current is character

Error (test-record_tests.R:25:3): covr.record_tests test traces list uses srcref key names
Error in `expect_match(names(attr(cov_func, "tests")), "\\w+(:\\d+){4,8}")`: is.character(act$val) is not TRUE
Backtrace:
 1. testthat::expect_match(names(attr(cov_func, "tests")), "\\w+(:\\d+){4,8}") test-record_tests.R:25:2
 2. base::stopifnot(is.character(act$val)) /Users/kirill/git/R/testthat/R/expectations-matches.R:35:2
```

krlmlr:
This occurs an M1 Mac.

dgkf:
Thanks @krlmlr. I don't have a physical M1 machine to test against, but it looks like there are a few web hosted offerings that I might be able to look into to do some investigation. I'll try my best to reproduce a similar setup to investigate where things are going wrong.

dgkf:
Sorry for a bit of a delay - it took me a while to hunt down an M1 system to test against. 

Initially, I was able to see the exact same tests failing, but after rebuilding and reinstalling they seem to disappear. I was only able to get these exact test failures if I installed the version of `covr` from `CRAN` before running the test suite. @krlmlr, could you rerun the test suite with a fresh install from source and see if the tests still fail?

Issue #448, opened 2020-09-15
compiled code not showing up on windows


tdhock:
hi @jimhester first of all thanks for maintaining covr which I find to be a very useful tool for many projects.

On my windows box the compiled code coverage is not showing up on the local report created via
```r
covr::report(covr::package_coverage(type="all"))
```
This is with https://github.com/tdhock/binsegRcpp -- I only see one file R/binseg_normal.R on the coverage report, and I expected to also see src/binseg_normal.cpp and src/interface.cpp as on codecov e.g. https://coveralls.io/builds/33368011

I do have gcov on the path, is there any other setup that I need to do to get this to work?
```
> system("gcov --version")
gcov (MinGW.org GCC-6.3.0-1) 6.3.0
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or 
FITNESS FOR A PARTICULAR PURPOSE.

[1] 0
> sessionInfo()
R version 4.0.2 (2020-06-22)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17134)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] htmlwidgets_1.5.1 compiler_4.0.2    R6_2.4.1          magrittr_1.5     
 [5] lazyeval_0.2.2    htmltools_0.5.0   tools_4.0.2       rex_1.2.0        
 [9] withr_2.2.0       yaml_2.2.1        DT_0.15           crosstalk_1.1.0.1
[13] jsonlite_1.7.1    digest_0.6.25     covr_3.5.0        rlang_0.4.7      
> str(covr::package_coverage(type="all"))
List of 18
 $ binseg_normal.R:24:3:24:44:3:44:34:34    :List of 3
  ..$ value    : num 3
  ..$ srcref   : 'srcref' int [1:8] 24 3 24 44 3 44 34 34
  .. ..- attr(*, "srcfile")=Classes 'srcfilealias', 'srcfile' <environment: 0x0000000017b6ac10> 
  ..$ functions: chr "binseg_normal"
[...]

PR #465, opened 2021-03-15
Add warnings for empty gcov (#448)

stephematician:
Add and test warnings if gcov missing or empty output to address #448

Simple warnings produced when there are .gcno files, but either gcov not found (i.e. covr.gcov is empty) or gcov output is empty.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 * `as.data.frame()` now returns an 0 row data frame when there are no functions in a package (#427)
 
 * Fix `parse_gcov` bug when package is stored in directory with regex special characters, see #459
+* Error/warning thrown for, respectively, missing gcov or empty parsed gcov output (@stephematician, #448)
 
 # covr 3.5.1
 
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks! I think this would be better off as an error rather than a warning, could you make that change and could you also please add a bullet t NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

stephematician:
Can do - but first I'd like to confirm. There are two different warnings I've inserted; should both be made into errors? Also I had a look at `icc.R` and similar checks there only result in warnings. Does this all make sense?

jimhester:
https://github.com/r-lib/covr/pull/465/files#diff-f748633604ad9005c853f36c4be958c6353e3ecb46bb4b3ee85bc4e7c3430669R73 should be an error I think.

stephematician:
Ok, yep that makes sense to me. I've switched the warning to an error (and updated corresponding test), + added to NEWS.md. Hopefully that covrs everything! ;) ... except for the pesky digest package that doesn't support 3.2

jimhester:
Thanks a bunch!

Issue #451, opened 2020-09-28
code coverage upload fails with Error in readRDS(files[i]) : error reading from connection


ajabadi:
### Observed behaviour

Uploading the code coverage succeeds on the fork but fails on main repo (same commit) with:

> Run covr::codecov(type = c('tests'))
> Error: Error in readRDS(files[i]) : error reading from connection
> Calls: <Anonymous> -> package_coverage -> merge_coverage
> Execution halted
> Error: Process completed with exit code 1.

Both used macOS-latest. Both public and no token used. Both repos added on codecov.io.

### Expected behaviour

When both the fork and main repo are (public and) added to codecov.io, we should get separate coverage reports for a push to both (while under the hood they could be the same report)

the fork Action: https://github.com/ajabadi/mixOmics/runs/1174435842
main repo action: https://github.com/mixOmicsTeam/mixOmics/runs/1174532979

the fork on codecov: https://codecov.io/gh/ajabadi/mixOmics/commits
main repo on codecov: https://codecov.io/gh/mixOmicsTeam/mixOmics

ajabadi:
Update: It seems to behave randomly on re-runs. Sometimes running successfully while other times failing.

jimhester:
covr writes the coverage information when the R process (or any child R process) exits, via a exit finalizer using `reg.finalizer(onexit = TRUE)`

I think likely something is killing the R process before the coverage information can be fully written, so the Rds files are being corrupted. Sometimes the data finishes writing before getting interrupted and it works, but other times it does not and fails.

I would look through your code and its dependencies for timeouts or `SIGKILL` signals and make sure they are set high enough for the data to be written fully and preferably use `SIGINT` so the process can exit cleanly.

ajabadi:
Thanks @jimhester 

I had followed the example in https://github.com/r-lib/actions/tree/master/examples#test-coverage-workflow
For some reason, it happens on `macOS-latest` but not on `ubuntu-latest` (on which it runs just fine so I use the latter to upload the coverage files). 

Also in later runs, it was not just an issue on the main repo but also on the fork. I'm not sure but my wild guess would be that it was caused by some recent changes to `macOS-latest`.

Issue #315, opened 2018-05-04
Error in readRDS(files[i]) : error reading from connection


wlandau:
On Travis, my tests pass, but `codecov()` fails: https://travis-ci.org/ropensci/drake/builds/374837082#L1293.

jimhester:
Could you run `covr::package_coverage()` locally to try and reproduce this?

wlandau:
Locally, `covr::package_coverage("drake")` runs normally. No errors, warnings, or messages. My session info is the same as for #314.

wlandau:
The problem went away in further commits to my repo. I do not know why. Closing.

randy3k:
I see the same error in Github Actions: https://github.com/REditorSupport/languageserver/runs/287088810

randy3k:
On the next build, the same error was seen on macOS but not the linux as above. The error was popping out randomly, not sure how it happens.

I was also able to reproduce it locally once.

randy3k:
I may have figured out the culprit. `callr::r_bg` was used in my package and it seems that the background process sometimes was killed when it was writing the rds file and left a corrupted file. It also explains the random nature of the error. Checking the process status carefully before quiting seems to fix the issue.

PR #471, opened 2021-04-16
Don't fail if rds files are corrupted

richfitz:
This is a somewhat unsatisfactory PR as I can't replicate this reliably.

As seen in https://github.com/r-lib/covr/issues/451 and https://github.com/r-lib/covr/issues/315 if the R process is killed when writing the Rds file it will create a corrupted rds file that will break the `covr::package_coverage()`. I've seen this rarely on gha builds (e.g. https://github.com/mrc-ide/rrq/pull/40/checks?check_run_id=2363889569) when using `callr::r_bg` fairly extensively, but have been unable to trigger in a MWE.

The downside is that the coverage report may be somewhat partial (and a message could be added during reading to indicate that) but continuing despite failure might be more graceful for relatively little additional complexity.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -468,7 +468,13 @@ merge_coverage <- function(files) {
 
   names <- names(x)
   for (i in 2:nfiles) {
-    y <- suppressWarnings(readRDS(files[i]))
+    # Avoid rare corrupted rds files from R processes that crash while
+    # writing coverage reports.
+    y <- tryCatch(suppressWarnings(readRDS(files[i])),
+                  error = function(e) NULL)
+    if (is.null(y)) {
+      next
+    }
     for (name in intersect(names, names(y))) {
       x[[name]]$value <- x[[name]]$value + y[[name]]$value
     }


jimhester:
I am pretty sure this is due to processx using SIGKILL in a number of places and the finalizer that is writing the coverage is interrupted.

@gaborcsardi would know better than I if there is a way to have the improved behavior for this use case, the current behavior is unfortunate.

gaborcsardi:
Sometimes you have to use sigkill, because R in the subprocess is not interruptible.

jimhester:
Sure, but if the process is writing the coverage out in an exit finalizer the process is already in the process of exiting, so there must be some case where the SIGKILL is too eager.

gaborcsardi:
@richfitz You can add some `$signal()` calls to your event loop layer if you want a graceful shutdown. processx is a bit too low level to implement this, without a standard event loop.

Or, if you don't mind sequential shutdown, then you can probably add a finalizer to your processes, that sends a `SIGTERM` first and then after a timeout a `SIGKILL`.

On Windows there is no `SIGTERM`, but you can `$interrupt()` first.

richfitz:
> @richfitz You can add some $signal() calls to your event loop layer if you want a graceful shutdown. processx is a bit too low level to implement this, without a standard event loop.

I don't think that this is the high level parts, unfortunately - I think that this is probably where a task has been spawned and is later cleaned up? I can't easily replicate unfortunately, and after replacing all the `$kill` calls I had (which I see uses `SIGKILL` after `SIGTERM` + grace period) I am still seeing intermittent failures - perhaps 1/3 to 1/4 of the time, so enough to be hard to work with but not enough to know if random fixes are helping!

I've not seen anything running `covr::package_coverage()` locally so I expect this is an issue with the speed at which the process is closed in the context of fairly slow GHA runners?

I do think that the current PR as it stands may be of use, but as always up to you for if you agree :)

PR #479, opened 2021-05-28
[hotfix] Nested coverage execution using covr.record_tests=TRUE (#463)

dgkf:
In using the `covr.record_test` feature added in #463 for testing, I discovered that it's possible to end up in a scenario where `covr` is tracing `covr:::count` calls from multiple packages by running `package_coverage` within a testing suite. In the case where the outer package's tests were not being record, but the inner package's tests are, then these `count` calls from the outer package will attempt to record tests despite not having the necessary fields in the `.counters[[key]]`. 

This fix adds a simple check for whether these fields are present before trying to append to them, which fixes the issues I've encountered while trying to run nested coverage runs for testing. 

Additionally, I've rebuilt the docs which added my name to the package Rd.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Encoding: UTF-8
 Package: covr
 Title: Test Coverage for Packages
-Version: 3.5.1.9001
+Version: 3.5.1.9002
 Authors@R: c(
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks!

PR #484, opened 2021-06-22
Ignore srcrefs without a file name

hadley:
Not sure if it's worth including this, but since I figured it out I thought I might as well record it.  In my case, the function was stored in `sysdata.rda`, so seems unlikely to be a common problem.


jimhester:
Thanks!

PR #485, opened 2021-07-22
Hotfix: Fix regression in how test depth is calculated; add exec index

dgkf:
This is a little bit of a dual part PR - first a fix to a regression introduced over the course of preparing #463, and the other piece is a small feature enhancement that enables a lot of powerful functionality.

### Depth calculation fix

Between the initial PR for #463 to the final commit, I went from just recording the final test frame to instead recording the entire test call stack up until the first `covr::count` call. Unfortunately, I forgot to update the way that the stack depth of the test trace was calculated. This has been fixed now by taking the length of the stack trace, instead of (what was) the index of the frame to begin on.

```diff
-  depth_into_pkg <- length(sys.calls()) - .current_test$frame - n_calls_into_covr + 1L
+  depth_into_pkg <- length(sys.calls()) - length(.current_test$frames) - n_calls_into_covr + 1L
```

Unfortunately, this was silently causing incorrect depth because of `rbind`'s value recycling. (effectively doing the same as `rbind(c(1, 2), c(3, NULL))`) ([location in diff](https://github.com/r-lib/covr/compare/master...dgkf:hotfix/test-trace-depth?expand=1#diff-d9130936e59a1de9d8ecb0695b16876ae6557ca3a23c80213d24d4574e07cf89R107)).

All is good now, and I added in some tests to avoid this from happening in the future.

### Feature Addition: Test execution index

As well, I added in an index for each trace. I'm pretty excited to have this introduced. Although it's just a tiny bit of added information, it effectively means that we can reconstruct the execution path that a test took while evaluating which opens up a lot of opportunities for analyzing the test traces. 

In practice we get something like this (below), where `i` records the order in which each trace is executed by each test. Combining this matrix across traces and sorting by "test" and "i" would let us see the order of coverage counts executed by a test. 

```
#' # $`source.Ref2326138c55:3:8:3:8:8:8:3:3`
#' #      test depth i
#' # [1,]    1     1 1
#' # [2,]    1     2 3
```

Materially, this means storing an extra column to the tests matrix for each trace. It comes at no substantial computational or object size cost.

---

- [x] Updated `?covr.record_tests` documentation to describe new data column
- [x] Added tests
- [x] Bumped DESCRIPTION dev version
- [x] Added reference to PR in NEWS entry regarding these changes

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Encoding: UTF-8
 Package: covr
 Title: Test Coverage for Packages
-Version: 3.5.1.9002
+Version: 3.5.1.9003
 Authors@R: c(
     person("Jim", "Hester", email = "james.f.hester@gmail.com", role = c("aut", "cre")),
     person("Willem", "Ligtenberg", role = "ctb"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 * Added `covr.record_tests` option. When `TRUE`, this enables the recording of
   the trace of the tests being executed and adds an itemization of which tests
   result in the execution of each trace. For more details see
-  `?covr.record_tests` (@dgkf, #463)
+  `?covr.record_tests` (@dgkf, #463, #485)
 
 * `package_coverage()` now sets the environment variable `R_TESTS` to the tests-startup.R file like R CMD check does (#420)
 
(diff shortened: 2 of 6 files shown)


jimhester (approved):

jimhester:
LGTM, is this ready to merge then?

dgkf:
Yes - it's good to go. Thanks @jimhester

jimhester:
Hey! Thanks!

PR #495, opened 2022-01-01
Remove unused function

alistaire47:
I'm not sure where why `f1()` here was added; presumably it was for debugging? Anyway, it's not used, so probably best to remove it and save a few bytes.

I ran `R CMD check` locally on these changes, and compiled tests are failing, but they fail whether or not it's there, I think because of my compiler configuration. Presumably CI checks in properly-configured environments will be more useful.


jimhester:
Thanks!

PR #496, opened 2022-01-03
make sure to only use functions of the same file in `to_sonarqube`

seasick:
When creating a SonarQube Generic XML from package coverage it can happen that the result contains wrong line numbers.

This can happen when you have multiple files with each of them containing an R6 class and each of them having an `initialize` function (or any other functions which are named the same).

This PR fixes that.


jimhester:
Thanks!

PR #499, opened 2022-01-19
Support strings in relative_path argument

### LINKED ISSUE #497: Add an option to report source-root-relative paths
`covr::package_coverage()` uses paths relative to the root of the R package by default, or absolute paths with `relative = FALSE` . None of these options is great for repositories where the R package sits underneath some subdirectory (duckdb, arrow, igraph, ...) . The problem occurs when trying to upload these coverage reports to codecov or other services. See e.g. https://codecov.io/gh/duckdb/duckdb/compare/624fa8aa4d81f812d7dc8d83620f914a76b3b114...e675b3dce58b120ca87574736f4c2386c1ddbcce/changes: this PR adds coverage checks for the R package that lives in `tools/rpkg` (via `covr::codecov("tools/rpkg")` but the files are not referred to by their relative root. This causes confusion with the other coverage checks in the project.

I propose supporting a third option, e.g. `relative = "wd"` or perhaps `relative = getwd()` where all paths would be stored relative to the working directory. Fixing this here is way easier than working around at the user level, it is very cumbersome to tweak the data structure returned by `covr::package_coverage()` .

krlmlr:
Closes #497.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -108,10 +108,10 @@ save_trace <- function(directory) {
 }
 
 #' Convert a counters object to a coverage object
-#' 
+#'
 #' @param counters An environment of covr trace results to convert to a coverage
 #'   object. If `counters` is not provided, the `covr` namespace value
-#'   `.counters` is used. 
+#'   `.counters` is used.
 #' @param ... Additional attributes to include with the coverage object.
 #'
 as_coverage <- function(counters = NULL, ...) {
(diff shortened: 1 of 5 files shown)


krlmlr:
This works well for duckdb, and might work for arrow and igraph and other packages too, I couldn't get codecov's path fixing to work. Would you support this feature?

jimhester:
I feel like codecov's path fixing should address this issue, but if you can't get it to work it is surely too difficult for most users to rely on.

Thanks for the PR! Hopefully this will be an easier API for people to use.

krlmlr:
Thanks for your support!

PR #510, opened 2022-06-24
Significant performance improvements for test recording (`options(covr.record_tests = TRUE)`)

dgkf:
The impetus for these changes was a staggering performance hit when recording tests while running coverage against the `rlang` package. The low-level nature of the package meant that some traces were hit millions of times with very deep call stacks - a perfect stress test for this feature. 

Compared to vanilla coverage, I was originally seeing ~70x slowdown with this feature enabled. With these changes, I've brought that down to ~4x. Still not negligible, but much, much better.

Functionally, changes are pretty minor. A slight data structure rework and some reworks to minimize call stack unwinding where possible. Full details below, but I'll also annotate the major changes in the patch.

### Changes

A number of changes were introduced to reduce the performance overhead of recording tests as they hit traces. Three main changes were introduced. These were discovered primarily during the testing of the `rlang` package which is a good stress test for low-level behaviors.

1. `sys.nframe` and `sys.call` were used in place of `sys.calls` where possible, avoiding unnecessary unwrapping of the call stack. For packages whose tests result in deep call stacks, repeated `sys.calls` calls was a major bottleneck.

2. A single value `.current_test$record` is used instead of repeatedly extracting the setting from `getOption("covr.record_tests")`. Avoiding calls to `getOption` with each `count` call limits the performance impact on vanilla `covr`.

3. `.counters[[key]]$tests` was converted from a matrix to an environment containing a matrix to avoid reallocation on assignment. This results in a substantial (~5x) reduction in performance overhead.

4. `.counters[[key]]$tests$tally` (the matrix logging test hits) now initializes with 4 NA_integer_ records and doubles in size each time the available records are filled. This causes a substantial reduction in allocations during testing. For example: `rlang` hits some lines of code over 1M times during testing, which would reallocate a matrix as many times.  
    
    Because of this restructuring of the tests data, some of the consumers of this data were also refactored to handle this structure. In the spirit of keeping the `covr.record_tests` features maintainable, I also refactored many of the non-vanilla code branches into separate functions.


R/covr.R
```diff
@@ -588,21 +608,16 @@ merge_coverage.list <- function(coverage_objs) {
...
-    test_idx <- match(names(y$tests), Filter(nchar, names(x$tests)))
-    new_test_idx <- if (!length(test_idx)) seq_along(y$tests) else which(is.na(test_idx))
-    test_idx[new_test_idx] <- length(x$tests) + seq_along(new_test_idx)
 
-    # append any tests that we haven't encountered in previous objects
-    x$tests <- append(x$tests, y$tests[new_test_idx])
-    y$tests <- NULL
+    # only affects coverage produced with options(covr.record_tests = TRUE)
+    clean_coverage_tests(y)
+    merge_coverage_tests(from = y, into = x)  # modifies environments 'x' and 'y' in-place
```
dgkf, line 618: In the spirit of keeping the "vanilla" `covr` workflow clean, I tried to refactor out a lot of the test recording behaviors into separate functions. This is a bit of a theme with some of the changes in this PR.

R/trace_calls.R
```diff
@@ -113,15 +113,7 @@ new_counter <- function(src_ref, parent_functions) {
...
-
-  if (isTRUE(getOption("covr.record_tests", FALSE))) {
-    .counters[[key]]$tests <- matrix(
-      numeric(0L),
-      ncol = 3L,
-      # test index; call stack depth of covr:::count; execution order index
-      dimnames = list(c(), c("test", "depth", "i")))
-  }
-
+  if (isTRUE(getOption("covr.record_tests", FALSE))) new_test_counter(key)
```
dgkf, line 116: Another example of cleaning up the "vanilla" path. The block that was removed was moved to be handled in `new_test_counter(key)`.

R/trace_calls.R
```diff
@@ -131,7 +123,7 @@ new_counter <- function(src_ref, parent_functions) {
 #' @keywords internal
 count <- function(key) {
   .counters[[key]]$value <- .counters[[key]]$value + 1L
-  if (isTRUE(getOption("covr.record_tests"))) count_test(key)
+  if (isTRUE(.current_test$record)) count_test(key)
```
dgkf, line 126: When profiling `covr`, I found that calls to `getOption` was introducing non-negligible overhead to the "vanilla" `covr` path. Storing the option when the counters are reset and using the value more directly cuts this down quite a lot, speeding up vanilla `covr` by ~5% for this `rlang` stress test. 

Most packages won't hit `count` nearly as often as `rlang` does, but for packages that do, this is a meaningful performance gain.

R/trace_tests.R
```diff
@@ -97,17 +97,60 @@ NULL
...
-  .counters[[key]]$tests <- rbind(
-    .counters[[key]]$tests,
-    c(length(.counters$tests), depth_into_pkg, .current_test$i)
+
+  # expand infrequently as new tests are added, doubling matrix size as needed
+  tests <- .counters[[key]]$tests
+  n <- NROW(tests$tally)
+  if (.counters[[key]]$value > n) {
+    tests$tally <- rbind(tests$tally, matrix(NA_integer_, ncol = 3L, nrow = n))
+  }
```
dgkf, line 114: Repeated calls to `rbind` to add a single row to a matrix were causing substantial slowdown due to reallocation with each `rbind`. Instead, the matrix now starts with 4 rows and doubles in size each time it is filled. When merging all the traces together, any unused rows are filtered out (see `clean_coveage_tests` & `merge_coverage_tests`)

R/trace_tests.R
```diff
@@ -97,17 +97,60 @@ NULL
...
+  tests$.data[[1L]] <- length(.counters$tests)
+
+  # call stack depth when trace is hit
+  tests$.data[[2L]] <- sys.nframe() - length(.current_test$frames) - n_calls_into_covr + 1L
+
+  # number of traces hit by the test so far
+  tests$.data[[3L]] <- .current_test$i
+
+  tests$.value <- .counters[[key]]$value
+  with(tests, tally[.value,] <- .data)
```
dgkf, line 126: The nested indexed assignment (`.counters[[key]]$tests <- ...`) was also resulting in a big performance hit due to reallocations. 

Instead, `.counters[[key]]$tests` is now an `environment` so that `.counters[[key]]$tests$tally` can be modified in place and avoid drilling down into the counters object.

R/trace_tests.R
jimhester: Yeah, repeated calls to rbind is basically the worst case performance pitfall in R.

Did you try storing the intermediate results in a list and then combining them in one pass with `do.call(rbind.list)` or faster equivalents like `rlang::vec_rbind()`? Generally those types of solutions end up being the fastest solution to this problem.

dgkf: Thanks @jimhester - I'll try that method out and report back. That would save some of the complexity of dealing with partially used matrices.

dgkf: @jimhester - I tried out a few approaches using `list`s or `environment`s to collect individual rows and then collecting them into a single matrix using `do.call(rbind, <list>)`. I wasn't able to outperform this implementation. I didn't try using the `rlang` alternatives because I don't think this behavior warrants the introduction of a new dependency. 

Assigning into a `.counters[[key]]$tests` `list` (eg, `.counters[[key]]$tests[[n]] <- val`) hit the same performance issues with nested listed assignment that I was running up against before. To get comparable performance, I still had to wrap it in an environment to avoid nested assignment which means the approach still has the same complexity in how the data is structured. 

Overall, I think the approach in the initial PR is probably still the best option.

R/trace_tests.R
```diff
@@ -97,17 +97,60 @@ NULL
...
+  .counters[[key]]$tests$.data <- vector("integer", 3L)
+  .counters[[key]]$tests$.value <- integer(1L)
+  .counters[[key]]$tests$tally <- matrix(
+    0L,
+    ncol = 3L,
+    # initialize with 4 empty rows, only expanded once populated
+    nrow = 4L,
+    # cols: test index; call stack depth of covr:::count; execution order index
+    dimnames = list(c(), c("test", "depth", "i"))
   )
```
dgkf, line 154: The matrix that is used for recording the test index each time a trace is hit is initialized with 4 rows. This was set based only on intuition. If it's ever filled, the matrix is extended to double its size. 4 initial rows was chosen as a balance between reallocation frequency and object size.

R/trace_tests.R
```diff
@@ -227,18 +272,10 @@ truncate_call <- function(call_obj, limit = 1e4) {
...
-    getSrcref(syscalls[[.current_test$last_frame]]) %||% syscalls[[.current_test$last_frame]]
-  ) ||
-  !identical(
-    .current_test$trace,
-    syscalls[.current_test$frames]
-  )
+  .current_test$last_frame > sys.nframe() ||
+  !identical(.current_test$src_call, sys.call(which = .current_test$last_frame)) ||
+  !identical(.current_test$src_env, sys.frame(which = .current_test$last_frame))
 }
```
dgkf, line 279: The major focus of this refactor is to reduce `sys.calls()` calls. The last call and environment in the call stack before we enter package namespace code is used as a heuristic to see if the test has changed. This is a bit less precise than when it was comparing the full call stack before, but is much, much faster. 

Situations where the same call, from the same environment actually corresponds to two different tests is possible, but highly unlikely. In my opinion, not worth the performance overhead of unwinding the call stack with each call to `covr::count()`.

dgkf:
I'm happy with where this is at now. Commit history got a bit garbled along the way, so changes have been squashed and rebased onto main. 

Changes were tested against `rlang`, and ran the test suite for [`covtracer`](https://github.com/Genentech/covtracer) - a tool that tightly integrates with these features, producing only expected test failures given the changes that were made (the change now uses `integer` matrices instead of `double` matrices). 

There is one failing action, but it seems to be expected.

jimhester:
Thanks, it all looks good!

PR #511, opened 2022-07-08
Hotfix (covr.record_tests): Handling case when 0 test expressions are hit

dgkf:
A small patch to fix an edge-case behavior that I hadn't considered in #510. 

When no tests were hit, the element that _would_ contain a list of tests remains `NULL` and the coverage object was previously assumed to have been produced without recorded tests. Now, a more direct measure of whether the feature was enabled is used, inspecting trace counters directly.

A test was also added to avoid regressions for this behavior going forward.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -642,7 +642,8 @@ merge_coverage.list <- function(coverage_objs) {
 #   may have been allocated, but not entirely populated.
 #
 clean_coverage_tests <- function(obj) {
-  if (is.null(obj$tests)) return()
+  counter_has_tests_tally <- function(counter) !is.null(counter$tests)
+  if (is.na(Position(counter_has_tests_tally, obj))) return()
 
   for (i in seq_along(obj)) {
     if (is.null(val <- obj[[i]]$value)) next
--- a/R/trace_tests.R
+++ b/R/trace_tests.R
@@ -145,7 +145,7 @@ new_test_counter <- function(key) {
   .counters[[key]]$tests$.data <- vector("integer", 3L)
   .counters[[key]]$tests$.value <- integer(1L)
   .counters[[key]]$tests$tally <- matrix(
-    0L,
+    NA_integer_,
     ncol = 3L,
     # initialize with 4 empty rows, only expanded once populated
     nrow = 4L,
--- a/R/trace_tests.R
+++ b/R/trace_tests.R
@@ -190,7 +190,10 @@ update_current_test <- function() {
 
   has_srcfile <- viapply(syscall_srcfile, length) > 0L
   srcfile_tmp <- logical(length(has_srcfile))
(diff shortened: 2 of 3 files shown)


R/covr.R
```diff
@@ -642,7 +642,8 @@ merge_coverage.list <- function(coverage_objs) {
 #   may have been allocated, but not entirely populated.
 #
 clean_coverage_tests <- function(obj) {
-  if (is.null(obj$tests)) return()
+  counter_has_tests_tally <- function(counter) !is.null(counter$tests)
+  if (is.na(Position(counter_has_tests_tally, obj))) return()
```
dgkf, line 646: `obj$tests` being `NULL` was not a sufficient way of determining whether counters would have test matrices. Previously, this would mean that matrices would be left with their pre-allocated records. 

Instead, using the existence of the `obj[[key]]$tests` element directly, this is a more reliable way of determining whether tests were captured.

R/trace_tests.R
```diff
@@ -145,7 +145,7 @@ new_test_counter <- function(key) {
   .counters[[key]]$tests$.data <- vector("integer", 3L)
   .counters[[key]]$tests$.value <- integer(1L)
   .counters[[key]]$tests$tally <- matrix(
-    0L,
+    NA_integer_,
```
dgkf, line 148: An minor, unrelated change. Using `NA` as a default value instead of `0`. Functionally, this has no impact in almost all cases. I've only seen one example where this value might have any effect, which is when testing interrupts. I found one instance where a counter's value was incremented, but the test hit wasn't recorded before the interrupt took effect.

R/trace_tests.R
```diff
@@ -190,7 +190,10 @@ update_current_test <- function() {
 
   has_srcfile <- viapply(syscall_srcfile, length) > 0L
   srcfile_tmp <- logical(length(has_srcfile))
-  srcfile_tmp[has_srcfile] <- startsWith(syscall_srcfile[has_srcfile], normalizePath(.libPaths()[[1]]))
+  srcfile_tmp[has_srcfile] <- startsWith(
+    syscall_srcfile[has_srcfile],
+    normalizePath(.libPaths()[[1]], mustWork = FALSE)
+  )
```
dgkf, line 196: Another sort of tangential bugfix. I found a situation where `normalizePath` would throw a "file or directory not found" error. I haven't discovered exactly what testing scenario somehow led to an invalid `.libPaths()`, but this is a simple hotfix just to safeguard against covr introducing errors into the test execution.

dgkf:
@jimhester - I don't mean to nag, but just wanted to draw your attention to this PR if you hadn't seen it yet. Let me know if you'd prefer not to be tagged just for awareness.

jimhester:
Thanks, it looks good to me!

PR #530, opened 2023-03-31
Feature Enhancement: Minimize duplicated recorded tests

### LINKED ISSUE #528: `options(covr.record_tests = TRUE)`: tests being captured multiple times in for loops
In light of #527, I wanted to let you know that we've discovered a behavior in the last couple days related to the `options(cov.record_tests = TRUE)` that I'd like to fix. If we can squeeze a fix into this release that would be awesome.

The overall behavior isn't _wrong_, but can result in really big coverage objects with inflated `attr(<coverage obj>, "tests")` due to how we currently decide when to log a new test. In most cases, the bloat is negligible, but the [`fastmap` package](https://github.com/r-lib/fastmap) piqued our interest because it is a rather pathological case where the bloat increases the object size by a few orders of magnitude. This is enough to crash the containers we're running the evaluation in. 

I've already pinpointed what needs to change, but it will take me a day or so to make the fix. There is a simple case that is just a couple lines of code change, and a non-trivial case that would require a bit more machinery to avoid duplicate test records.

# What's causing the bloat?

## The simple case

When the same test is hit repeatedly in a loop, we are currently recording a "new" test for each call from the test code.

The issue lies in the heuristics we use to determine when we're onto a new test call. When tests are evaluated in a loop (for example, [this `for` loop from `fastmap`](https://github.com/r-lib/fastmap/blob/147830b6e691cc5f2da44f6c956f5164e122d75a/tests/testthat/test-map.R#L274-L292)), we record a new test each time the evaluation frame changes. Instead, we should only consider it as a new test if the _calling_ frame changes. This is a simple fix, and just means we need to take the `parent.env()` instead of the `sys.frame` itself.

Fixing this is just a matter of changing a couple lines, and [is already implemented](https://github.com/dgkf/covr/commit/67d511aeabb9ac3c17f9569a5a6e3b328746c868).

## The non-trivial case

This is a bit more complicated when we _alternate_ calls into a package within a loop (as seen in [this `for` loop from `fastmap`](https://github.com/r-lib/fastmap/blob/147830b6e691cc5f2da44f6c956f5164e122d75a/tests/testthat/test-queue.R#L243-L256), alternating between `q$remove` and `q$add`). In this case the call is changing and our heuristic is working as intended, but because we only check to see if the _last_ test is the same, each alternating call is still recorded.

To address this, I'd like to build a vector of test hashes (probably just a srcref string, ie `file:line:col`) of test calls that we can reference to see if the current test has already been logged and add a new column to each trace's test log that includes a record of how many times each test call was called. Proposed change to documented structure would look like

https://github.com/r-lib/covr/blob/caed186fc47884d1e848c09ddb9433716bd671c6/R/trace_tests.R#L80-L83
```diff
  #' # $`source.Ref2326138c55:3:8:3:8:8:8:3:3`
- #' #      test depth i
- #' # [1,]    1     1 1
- #' # [2,]    1     2 3
+ #' #      test call depth i
+ #' # [1,]    1    1     1 1
+ #' # [2,]    2    1     1 1
+ #' # [3,]    1    2     2 3
+ #' # [4,]    2    2     2 3
```
 
 [I implemented this today](https://github.com/dgkf/covr/commit/aa538fcf4773d1b9b34e1de2e7a00f053b733699), but the code isn't to a level of quality that I'm happy with yet and I want to do a little refactoring still before contributing it.

dgkf:
Closes #528 

Refer to #528 for a detailed description of the issue, but in short the heuristic that was used for quickly determining whether we're evaluating a new test ended up being too specific, resulting in the same test being logged many times.

Generally this isn't an issue. Even in situations where a test suite might loop over a test, the heuristic would _usually_ treat those as the same test. Even if it was logged with each iteration, those iteration counts are usually low enough that it doesn't impact the coverage object substantially. However, when testing `fastmap`, there were a couple cases where an expectation is made in a loop with 10s of thousands of iterations and logging the test with each evaluation, blowing up the coverage object memory needs. 

To emphasize, these changes will only affect behaviors when `options(covr.record_tests = TRUE)`, which must be opted into. 

Technical details annotated as in-line comments below:


R/trace_tests.R
```diff
@@ -68,23 +69,23 @@
...
+#' # [1,]    1    1     2 4
 #'
 #' # reconstruct the code path of a test by ordering test traces by [,"i"]
 #' lapply(cov, `[[`, "tests")
 #' # $`source.Ref2326138c55:4:6:4:10:6:10:4:4`
-#' #      test depth i
-#' # [1,]    1     1 2
+#' #      test call depth i
+#' # [1,]    1    1     1 2
 #' #
```
dgkf, line 80: A change to the structure of the `<coverage>[[<key>]]$tests` elements.

Adds a `call` column, which is the number of times the test expression was evaluated before hitting this trace. It's not used within `covr`, but since the same test now represents multiple calls it is useful for distinguishing test evaluations useful for downstream tooling.

R/trace_tests.R
```diff
@@ -214,38 +218,100 @@ update_current_test <- function() {
...
+    )
+  } else {
+    # otherwise we compare call stacks
+    Position(
+      function(t) identical(t[], .current_test$trace),  # t[] to ignore attr
+      .counters$tests,
+      right = TRUE,
+      nomatch = length(.counters$tests) + 1L
+    )
+  }
```
dgkf, line 297: New tests are likewise distinguished by either their srcref "key" if the test code has known srcrefs, or otherwise an identical call stack.

R/trace_tests.R
```diff
@@ -214,38 +218,100 @@ update_current_test <- function() {
...
+}
```
dgkf, line 298: If the current test matches an already logged test, then we reuse that index. Otherwise it's added to a growing list of recorded test expressions.

New tests are distinguished by comparing the srcref "key" to any previous test keys if the test code has known srcrefs, or otherwise looking for identical call stacks.

I was expecting call stack comparisons like this to come with a big performance hit, but it ended up being pretty minor.

PR #566, opened 2024-04-28
Add codecov token as Authorization header

gaborcsardi:
Hopefully global tokens work this way.


gaborcsardi:
Never mind, this does not work.

Issue #574, opened 2024-08-12
package_coverage() can fail if the package is already loaded


kevinushey:
On Windows, R will refuse to install a package if that package is already loaded, even if the package is being installed into some independent library paths. This error comes from here:

https://github.com/wch/r-source/blob/0b3e527dc50303c6e320deb7bb2d11a02d636b71/src/library/utils/R/packages2.R#L278-L295

And `install.packages()` gets invoked here:

https://github.com/r-lib/covr/blob/dd5286dd7cb3909456637a52d942f2428fa0a926/R/covr.R#L443-L455

This implies that a workflow like:

```
devtools::load_all()
devtools::test_coverage()
```

will fail, as the package we want to run coverage on is already loaded.

PR #575, opened 2024-08-12
invoke package installation in separate R process

kevinushey:
Closes https://github.com/r-lib/covr/issues/574.


jimhester:
Thanks!

Issue #567, opened 2024-05-01
segfault recently in codecov()


MichaelChirico:
We are observing a segfault in codecov recently, e.g.

https://github.com/Rdatatable/data.table/actions/runs/8885891976/job/24398250857?pr=6107

Unfortunately I haven't been able to reproduce it outside CI, but it's pretty persistent across all commits in the past 2 days or so:

https://github.com/Rdatatable/data.table/actions/workflows/test-coverage.yaml

The stack trace is into `try()`, but I am not sure how to proceed from here, any ideas?

mcol:
Your last green run on master was on 24 April, which is when R 4.4.0 was released. Something must have changed in that release that is negatively affecting `covr`. In https://github.com/R-Lum/Luminescence/issues/144 we've also attempted to debug our incredibly long coverage times, and we reached the conclusion that they started with R 4.4.0: we didn't see any segfault, but coverage times have increased by at least 10x. We're now running coverage on R 4.3.3 without issues.

struckma:
> Your last green run on master was on 24 April, which is when R 4.4.0 was released. Something must have changed in that release that is negatively affecting `covr`. In [R-Lum/Luminescence#144](https://github.com/R-Lum/Luminescence/issues/144) we've also attempted to debug our incredibly long coverage times, and we reached the conclusion that they started with R 4.4.0: we didn't see any segfault, but coverage times have increased by at least 10x. We're now running coverage on R 4.3.3 without issues.

Indeed, I have observed also some connection with the R version 4.4.0, but I could not really identify the reason. I did also dig in R's C-sources, but there is nothing really obvious. The NEWS.md mentions some changes around I/O, and there was a change in the serialization code.

I could not reliably reproduce the issue in a toy example, but it seems to be related to containerization, parallel test execution and maybe to the way, R tests are started (using the `code = {}` argument of `covr::package_coverage()`, e.g.

mcol:
I've bisected the R changes between 4.3.3 and 4.4.0 (using [wch's R-source mirror](https://github.com/wch/r-source)), and I identified [this commit](https://github.com/wch/r-source/commit/990fe3d8a10cca10e1181bd9ca41bbc8588b12f0) as the one where the problems started. This refers to [Bugzilla PR 16756](https://bugs.r-project.org/show_bug.cgi?id=16756), which indeed mentions `covr` as being related.

[...]

PR #577, opened 2024-09-25
Guess a fix to segfault issue

MichaelChirico:
For #567.

cc @mcol / @struckma -- can you test if this change removes the segfault for you? I'm still hamstrung by being unable to reproduce the segfault outside GHA.

Also cc @krlmlr who filed & followed the orginal Buganizer report & workaround kludge here.

### DIFF
--- a/R/parse_data.R
+++ b/R/parse_data.R
@@ -138,6 +138,10 @@ clean_parse_data <- function() {
 }
 
 # Needed to work around https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16756
-get_tokens <- function(srcref) {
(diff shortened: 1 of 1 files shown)


mcol:
I've run our testsuite with a freshly-compiled `covr` with your patch applied, and unfortunately it doesn't help: running time was 68 minutes (it's around 7-8 minutes on 4.3.3).

MichaelChirico:
Alas. IMO the next thing to do is try "bisecting" the package itself to see if it's a certain piece of code that's responsible. Unfortunately for data.table, the issue is most likely related to `` `[.data.table` `` which is ~2000 lines long and probably used in >80% of tests... not very friendly for bisection :\

PR #581, opened 2024-11-07
Message doesn't have a `call.` argument

hadley:


jimhester:
Oh yeah, apparently missed that when changing this from a `stop()` in https://github.com/r-lib/covr/commit/2981e49dc740539628b8ee57e4bb76775fd5d2f4

PR #584, opened 2024-11-07
Bump required R version to R 4.0.0

hadley:
(Since other tidyverse/r-lib packages will requiring it; this just makes the dependency explicit for covr)


jimhester:
I guess we should also stop testing on 3.6 in the CI along with this? https://github.com/r-lib/covr/actions/runs/11728979579/job/32673603649?pr=584

.github/workflows/test-coverage.yaml
```diff
@@ -15,40 +17,45 @@ jobs:
...
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
+          extra-packages: any::covr, any::xml2
           needs: coverage
 
-      - name: Setup
-        run: |
-          R CMD INSTALL .
-          source shim_package.sh
```
jimhester, line 31: This is critically important to run covr on itself, because you get into an infinite recursion issue if you don't rename the package to something else before trying to run covr on it, which is what shim_package.sh does.

.github/workflows/test-coverage.yaml
```diff
@@ -15,40 +17,45 @@ jobs:
...
+          cov <- covr::package_coverage(
             quiet = FALSE,
             clean = FALSE,
-            install_path = file.path(Sys.getenv("RUNNER_TEMP"), "package")
+            install_path = file.path(normalizePath(Sys.getenv("RUNNER_TEMP"), winslash = "/"), "package")
           )
+          covr::to_cobertura(cov)
         shell: Rscript {0}
 
+      - uses: codecov/codecov-action@v4
```
jimhester, line 41: What is the advantage of using the codecov-action vs having covr upload to codecov directly?

.github/workflows/test-coverage.yaml
hadley: Oops, I missed that.

.github/workflows/test-coverage.yaml
hadley: I don't know 😞 I just updated using our standard scripts. @gaborcsardi might know.

gaborcsardi: covr uses the old protocol, which does not support organization level codecov tokens.

jimhester: 🤔

gaborcsardi: Well, you don't have any problems if you don't look at the results. :) Upload failures do not fail the workflow run, so you typically do not notice them:
```
[1] "Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 1300s."
```
https://github.com/r-lib/covr/actions/runs/11729639825/job/32675733827#step:6:88

The last successful upload for covr was ~7 months ago.

jimhester: I seem to no longer have access to the admin settings for this repo, so can't comment on if codecov app is still installed for this repo or the r-lib org in general.

But I don't really like going through the cobutura codepath for repositories by default, it adds some additional complexity and also a xml2 dependency. Additionally the output is really geared towards more java style classes, and doesn't perfectly align with R package organization. Additionally doesn't relying on secrets mean that forks are no longer supported?

I am not sure why the existing path would work differently with organization tokens vs the action? Aren't they both changing the secret into an environment variable in the same way?

PR #580, opened 2024-10-29
Initial work on S7 support for covr

jimhester:
Fixes #578

PR #585, opened 2024-11-08
S7 external generics

t-kalinowski:
This builds on #580 by adding support for external generics registered via `` S7::`method<-` ``.


t-kalinowski:
Opened the same PR agaainst the S7 branch, for a cleaner diff. That PR should should automatically transfer over to main once #580 is merged.

PR #586, opened 2024-11-08
Add support for S7 methods registered for external generics

t-kalinowski:
This builds on #580 by adding support for external generics registered via `` S7::`method<-` ``.


jimhester:
Thanks, having good support for S7 should hopefully help with S7 adoption.

PR #587, opened 2024-11-12
Fix check NOTE: non-API calls to SET_BODY, SET_CLOENV, SET_FORMALS

t-kalinowski:
This PR is an unsavory but pragmatic and expedient solution to the current R CMD check NOTE 
https://cran.r-project.org/web/checks/check_results_covr.html
> ```
> Result: NOTE 
>   File ‘covr/libs/covr.so’:
>     Found non-API calls to R: ‘SET_BODY’, ‘SET_CLOENV’, ‘SET_FORMALS’
>   
>   Compiled code should not call non-API entry points in R.
> ```

This can be a “temporary” solution until the internal layout of a `CLOSXP` changes, closure modifiers become part of R’s C API, or we decide to undertake a larger refactor of covr to avoid needing this workaround.


jimhester:
I wonder if we could explain this use case to Luke Tierney and see if we could get a API function for this. There is [Rf_mkClosure](https://github.com/wch/r-source/blob/f50aab97c98615643db276bb034701816ab43143/src/main/dstruct.c#L116) which almost does what we need, but always allocates a new closure rather than swapping in place.

Making us dependent on the internal CLOSEXP memory layout seems like moving in the wrong direction from what R core intended with the extra checks.

Though if they are not going to change we may have to do this for now.

If we also captured the environment that the original function was stored in I guess we could just replace the function with the new version in that environment and avoid the in place swapping, but it would take some refactoring of covr to accomplish.

t-kalinowski:
> seems like moving in the wrong direction from what R core intended

That may be true. The usage in covr is unique, and I don’t anticipate an API endpoint to make this straightforward anytime soon. I understand and sympathize with R Core's desire to proceed cautiously and intentionally here. With this approach, the risks and assumptions are clearly defined and managed within covr, with the understanding that we’ll stay responsive to any upstream changes. I think that aligns somewhat with R Core’s goal of distinguishing between what's officially supported and what's not.


> If we also captured the environment that the original function was stored in

This would be a relatively involved refactoring, since not all functions are resolved via a binding in an environment. E.g., an S7 property setter is resolved from a list, and an S7 class validator from an attribute.

jimhester:
Can we add a note to news about this change, then I can merge it. I appreciate you looking into it and coming up with a solution, even if it is not ideal, we have to work with what CRAN allows us to do.

src/reassign.c
```diff
@@ -1,19 +1,62 @@
...
 #include <R.h>
+#include <Rinternals.h>
 #include <R_ext/Error.h>
 #include <R_ext/Rdynload.h>
 #include <Rdefines.h>
 #include <stdlib.h>  // for NULL
+#include <stdint.h> // for uint64_t
+
+// Mirror the exact structures of SEXPREC from R internals
+struct proxy_sxpinfo_struct {
```
hadley, line 11: I'd move the comments here just because this is the first thing that you read where you wonder what's going on.

src/reassign.c
```diff
@@ -1,19 +1,62 @@
...
-  SET_BODY(old_fun, BODY(new_fun));
-  SET_CLOENV(old_fun, CLOENV(new_fun));
+  // The goal is to modify `old_fun` in place, so that all existing references
+  // to `old_fun` call the tracing `new_fun` instead.
+  // This used to be simply:
+  //   SET_FORMALS(old_fun, FORMALS(new_fun));
+  //   SET_BODY(old_fun, BODY(new_fun));
+  //   SET_CLOENV(old_fun, CLOENV(new_fun));
+  // But those functions are now "non-API". So we comply with the letter of the law and
+  // swap the fields manually, making some hard assumptions about the underling memory
```
mcol, line 44:
```suggestion
  // swap the fields manually, making some hard assumptions about the underlying memory
```

src/reassign.c
```diff
@@ -1,19 +1,62 @@
...
+  // layout in the process. See also: "The Cobra Effect" (https://en.wikipedia.org/wiki/Cobra_effect).
+  // Rather than using memcpy() with a hard coded byte offset,
```
mcol, line 46:
```suggestion
  // Rather than using memcpy() with a hard-coded byte offset,
```

mcol (commented):
Just two typos in the comments.

src/reassign.c
t-kalinowski: Move the struct defs down so they're after the comment.

jimhester:
Thanks again for working on this!

PR #588, opened 2024-11-15
parse_data: Fix compatibility with R 4.4

### LINKED ISSUE #576: Error in exclude() 'from must be a finite number' for seq() usage
Same error as old #322, but not related; this is coming from a different stack trace entirely:

```
9: stop("'from' must be a finite number")
8: seq.default(df[i, "first_line"], df[i, "last_line"])
7: seq(df[i, "first_line"], df[i, "last_line"])
6: seq(df[i, "first_line"], df[i, "last_line"]) %in% excl[[file]]
5: FUN(X[[i]], ...)
4: vapply(X, FUN, ..., FUN.VALUE = logical(1))
3: vlapply(seq_len(NROW(df)), function(i) {
       file <- df[i, "full_name"]
       which_exclusion <- match(file, names(excl))
       !is.na(which_exclusion) && (identical(excl[[which_exclusion]], 
           Inf) || all(seq(df[i, "first_line"], df[i, "last_line"]) %in% 
           excl[[file]]))
   })
2: exclude(coverage, line_exclusions = line_exclusions, function_exclusions = function_exclusions, 
       path = root)
1: covr::package_coverage()
```

Most likely this is related to #567. I am trying to reproduce the CI error locally, to no avail.

I _will_ note that `package_coverage()` felt interminable -- the execution took at least 40 minutes. There was no segfault, though it did not succeed.

I'll try and debug the error but given the run times it's very hard to reproduce; filing first.

### LINKED ISSUE #579: Applying exclusions raise error due to missing srcref lines
Identified this odd bug when trying to run `covr` against `diffobj`. It's unclear to me why this is popping up for this package.

```r
library(covr)  # using main @c0c88f1
package_coverage("./diffobj")
```

Coverage tallying finishes successfully, but when applying exclusions will produce the error:

```
Error in seq.default(df[i, "first_line"], df[i, "last_line"]) :
  'from' must be a finite number
> traceback()
9: stop("'from' must be a finite number")
8: seq.default(df[i, "first_line"], df[i, "last_line"])
7: seq(df[i, "first_line"], df[i, "last_line"])
6: seq(df[i, "first_line"], df[i, "last_line"]) %in% excl[[file]] at exclusions.R#90
5: FUN(X[[i]], ...)
4: vapply(X, FUN, ..., FUN.VALUE = logical(1)) at utils.R#346
3: vlapply(seq_len(NROW(df)), function(i) {
       file <- df[i, "full_name"]
       which_exclusion <- match(file, names(excl))
       !is.na(which_exclusion) && (identical(excl[[which_exclusion]],
           Inf) || all(seq(df[i, "first_line"], df[i, "last_line"]) %in%
           excl[[file]]))
   }) at exclusions.R#85
2: exclude(coverage, line_exclusions = line_exclusions, function_exclusions = function_exclusions,
       path = root) at covr.R#544
1: covr::package_coverage("diffobj"
```

There's a single trace in the coverage trace data frame with a `NA` `first_line` (among other columns), and this can also be see in the pre-transformed `coverage` object. When debugging the coverage traces before applying exclusions, one trace contains:

```r
$`capt.R:NA:NA:NA:NA:NA:NA:NA:NA`
$`capt.R:NA:NA:NA:NA:NA:NA:NA:NA`$value
[1] 100

$`capt.R:NA:NA:NA:NA:NA:NA:NA:NA`$srcref
<srcref: file "diffobj/R/capt.R" chars NA:NA to NA:NA>

$`capt.R:NA:NA:NA:NA:NA:NA:NA:NA`$functions
[1] "capt_chr"
```

kyleam:
This series resolves two issues with calculating package coverage under R 4.4:

 * for some packages, there is a substantial performance regression compared to R 4.3 (related to gh-567)

 * some packages hit into an error when applying exclusions (gh-576, gh-579)

In addition to the linked issues, there are cases demonstrating both these issues at <https://github.com/kyleam/covr-r44-issues>.

Here's a summary of timings (seconds, wall time) to give a sense of how severe the slowdown can be:

| package    | R 4.3.3 | R 4.4.2 | R 4.4.2 + patch |
|----------- |-------- |-------- |---------------- |
| data.table | 254     | 2290    | 235             |
| diffobj    | 231     | 1126    | 151             |
| R.utils    | 759     | 22197   | 599             |

("patch" refers to a run with a covr patched with the equivalent of the changes proposed here.)

---

Both the slowdown and error bisect to R's [990fe3d8a1](https://github.com/wch/r-source/commit/990fe3d8a10cca10e1181bd9ca41bbc8588b12f0) (getParseData() - for first-file objects ..., 2023-06-14).  And the [associated bug report](https://bugs.r-project.org/show_bug.cgi?id=16756) was actually connected to a covr PR (gh-154).

As discussed in the third commit's message, I think covr is getting tripped up by the new getParseData  returning parse data for the entire package.

<details>
<summary>example</summary>

Under R 4.4.1, `getParseData` returns the same result for `nanaduration` and `nanoival`, which are in separate files.  The parse data appears to be for all files of the package.

```R
d1 <- getParseData(nanotime::nanoduration)
d2 <- getParseData(nanotime::nanoival)

basename(attr(d1, "srcfile")[["filename"]])
#> [1] "nanoduration.R"
basename(attr(d2, "srcfile")[["filename"]])
#> [1] "nanoival.R"

attr(d1, "srcfile") <- NULL
attr(d2, "srcfile") <- NULL

identical(d1, d2)
#> [1] TRUE
```

</details>

I haven't worked out the details of how that leads to the `NA` values that trigger exclusions error.

My proposed solution is to reorder the `utils::getParseData` and `covr:::get_parse_data` calls so that that `get_parse_data` is used for R 4.4 or greater, just as it is for R 4.3 or below.  As a result, the `get_tokens` call returns the expected per-file parse data.

The `utils::getParseData` call is retained as the fallback.  That's needed when not calculating *package* coverage (e.g., when calling `file_coverage`).

Putting `get_parse_data` in front of the `getParseData` call exposes  two issues in `split_on_line_directives`.  Those are resolved in the first two commits.

Closes #576
Closes #579

cc: @dgkf @jimhester @krlmlr @mcol @MichaelChirico


MichaelChirico:
I don't know the internals well enough to comment further than: looks promising!!! thank you 🙌

jimhester:
Your investigation makes sense thanks, do you think you could add a test for the `split_on_line_directives` cases that are fixed with this PR? Otherwise this looks good to me, appreciate the work!

kyleam:
@jimhester thanks for taking a look!

> do you think you could add a test for the split_on_line_directives cases that are fixed with this PR?

Sure, I've pushed an updated.  (The issues are triggered by existing tests, but I agree that an direct/explicit makes sense.)

mcol:
I've tried the patch and it brings the time for running coverage on Luminescence back to 7 minutes, which is roughly what we get on 4.3.3 (as opposed to 80m with the version on CRAN with 4.4.1). :tada:

jimhester:
Thanks again for the work and thank you @mcol for reporting on the improved runtime.

PR #589, opened 2024-12-12
Address testthat deprecation warnings + drop mockery

olivroy:
Fix #595

* Replace mockery with testthat new mocking approach
* Replace deprecated `testthat::with_mock()` with `testthat::with_mocked bindings()`. (Addresses warnings logged in https://github.com/r-lib/covr/actions/runs/12316159627/job/34375845219#step:6:185)

They turned more challenging to fix than I expected.

I therefore removed the mockery dependency and `testthat::with_mock()` usage, but that required a bit of refactoring, such as importing from `{httr}` namespace to be able to mock `httr::content`, `httr::RETRY`, etc.

Kept everything related to mocking here to address testthat 3.2.2 deprecation warnings and sent other things to other PRs.

PR #590, opened 2024-12-13
Misc cleanups

olivroy:
Branched out from #589 

This only contains trivial cleanups, fixing .lintr file and fix some trivial lints


jimhester:
Thanks, thee look good!

PR #591, opened 2024-12-13
Use cli instead of crayon for console output

olivroy:
Branched out from #589

The visual change is the following:

Current CRAN:

image

This PR:
image

Since covr doesn't import rlang, we can't use `cli::cli_inform()`, so I am using `cli::format_message()` + `message()` instead


jimhester (approved):

jimhester:
Thanks!

PR #593, opened 2025-01-07
Normalize install_path before creating dir

### LINKED ISSUE #592: `install_path` parameter breaks in WINDOWS subprocesses
When running the `covr` in the `callr` subprocess on Windows, `covr::package_coverage` unexpectedly breaks if `install_path` has double backslashes (valid WINDOWS path) `\\` instead of the single forward slashes `/`. That's due to the path normalization which happens after creating the directory: 

https://github.com/r-lib/covr/blob/ca4241da03a30b76a8d8cbfc194371cb94ca512e/R/covr.R#L413-L415

Swapping these two lines will make the internal `normalize_path` function first replace all the `\\` with `/` so the function can proceed as expected. I'm not sure why exactly this is an issue in a subprocess, but the change looks reasonable either way.

maksymiuks:
Closes #592 

Candidate fix to the issue reported on Windows machines.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,10 @@
 
 * Added support for `klmr/box` modules. This works best with `file_coverage()`. (@radbasa, #491)
 
+* Normalize `install_path` path before creating directory to prevent
+  failures when during running covr in subprocess a path with Windows 
+  `\\` file path separator is provided. (@maksymiuks, #592)
+
 # covr 3.6.4
 
 * Fix for a failing test on CRAN
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,8 +16,8 @@
 * Added support for `klmr/box` modules. This works best with `file_coverage()`. (@radbasa, #491)
 
 * Normalize `install_path` path before creating directory to prevent
-  failures when during running covr in subprocess a path with Windows 
-  `\\` file path separator is provided. (@maksymiuks, #592)
+  failures when running covr in a subprocess using a path with Windows 
+  `\\` path separators. (@maksymiuks, #592)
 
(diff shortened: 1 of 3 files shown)


R/covr.R
```diff
@@ -410,9 +410,8 @@ package_coverage <- function(path = ".",
     root <- NULL
   }
 
-  dir.create(install_path)
   # tools::testInstalledPackage requires normalized install_path (#517)
-  install_path <- normalize_path(install_path)
+  dir.create(install_path <- normalize_path(install_path))
```
jimhester, line 414:
```suggestion
  install_path <- normalize_path(install_path)
  dir.create(install_path)
```

NEWS.md
```diff
@@ -15,6 +15,10 @@
 
 * Added support for `klmr/box` modules. This works best with `file_coverage()`. (@radbasa, #491)
 
+* Normalize `install_path` path before creating directory to prevent
+  failures when during running covr in subprocess a path with Windows 
+  `\\` file path separator is provided. (@maksymiuks, #592)
```
jimhester, line 20:
```suggestion
* Normalize `install_path` path before creating directory to prevent
  failures when running covr in a subprocess using a path with Windows 
  `\\` path separators. (@maksymiuks, #592)
```

jimhester (changes requested):
Thanks, made a couple of suggested edits, otherwise looks good.

maksymiuks:
Thanks for the review, I've applied your suggestions :)

jimhester (approved):

jimhester:
Thanks!

PR #598, opened 2025-01-23
Further refinement of error for mismatched start/end

MichaelChirico:
In large files, this will help pinpoint where to look by visually inspecting which start/end looks unpaired


MichaelChirico:
FWIW, I know this is essentially a fork of the similar logic in {lintr} (or vice versa), here is the current state of the error there:

https://github.com/r-lib/lintr/blob/b6d945d980e726029f89f96d7124b74541d8ee9a/R/exclude.R#L135-L138

jimhester (approved):

PR #607, opened 2025-05-09
Better experience of failing package installation

MichaelChirico:
I have some C++ lib issue that was hidden by `quiet=TRUE`, I only experienced this as:

```
Error in file(con, "r") : cannot open the connection
In addition: Warning message:
In file(con, "r") :
  cannot open file '/tmp/RtmpryiHR0/R_LIBS25202924885e/TestCompiled/R/TestCompiled': No such file or directory
```

Not very helpful.

### DIFF
--- a/R/covr.R
+++ b/R/covr.R
@@ -785,8 +785,7 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE,
 
   trace_dir <- paste0("Sys.getenv(\"COVERAGE_DIR\", \"", lib, "\")")
 
-  load_script <- file.path(lib, pkg_name, "R", pkg_name)
-  lines <- readLines(file.path(lib, pkg_name, "R", pkg_name))
+  lines <- readLines(load_script)
   lines <- append(lines,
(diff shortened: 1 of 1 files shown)


R/covr.R
```diff
@@ -785,8 +789,7 @@ add_hooks <- function(pkg_name, lib, fix_mcexit = FALSE,
 
   trace_dir <- paste0("Sys.getenv(\"COVERAGE_DIR\", \"", lib, "\")")
 
-  load_script <- file.path(lib, pkg_name, "R", pkg_name)
-  lines <- readLines(file.path(lib, pkg_name, "R", pkg_name))
+  lines <- readLines(load_script)
```
MichaelChirico, line 792:
```suggestion
  load_script <- file.path(lib, pkg_name, "R", pkg_name)
  lines <- readLines(load_script)
```

jimhester:
I think we should probably make this an error rather than a warning. I don't know of any cases where a non-zero return from package installation would still be workable.

PR #608, opened 2025-05-09
Optionally make partial coverage "uncovered"

MichaelChirico:
Not well tested yet -- my machine is some sort of borked & I haven't had time to fix it.

Checking in now since I think it's already in a good place for at least high-level review.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # covr (development version)
 
+* Support for excluding partially-covered lines in compiled code (@MichaelChirico, #604).
+  Use `options(covr.gcov_exclude_partial_lines = TRUE)` to require every statement on a line be
+  executed before the line is counted a "covered". For example, `if (verbose) Rprintf("hi\n");`
+  would require a test under `verbose=true` in this stricter setting.
+
 * Messages are now displayed using cli instead of crayon (@olivroy, #591).
 
 * covr now uses `testthat::with_mocked_bindings()` for its internal testing (@olivroy, #595).
--- a/R/compiled.R
+++ b/R/compiled.R
@@ -16,7 +16,7 @@ parse_gcov <- function(file, package_path = "") {
   }
 
   re <- rex::rex(any_spaces,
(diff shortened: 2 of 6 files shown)


MichaelChirico:
OK, looks like there's some version dependency problems where `*` is not supported. Would help to have some experts weigh in :)

I see this on the failing `macos`:

```
Apple LLVM version 15.0.0
  Optimized build.
  Default target: arm64-apple-darwin23.6.0
  Host CPU: apple-m1
```

This on the failing `windows`:

```
gcov (x86_64-posix-seh, Built by MinGW-W64 project) 4.9.3
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or 
FITNESS FOR A PARTICULAR PURPOSE.
```

And this on the passing `linux`:

```
gcov (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

MichaelChirico:
It looks like `-a` can help and is present on all versions, it gives output like this on `macos`:

```
        1:   34:  if (REAL(x)[0] > 0) return Rf_ScalarLogical(TRUE);
        1:   34-block  0
    $$$$$:   34-block  1
```

And on `ubuntu`:

```
         1*:   34:  if (REAL(x)[0] > 0) return Rf_ScalarLogical(TRUE);
          1:   34-block  0
      %%%%%:   34-block  1
```

But that again is a bit of overkill -- requires refactoring how the `gcov` output is processed to involve _aggregating_ from the `(line, block)` to the `(line)` level. I'm also not familiar enough with if there are other implications to work through.

For now, I'll orient on a `has_partial_coverage()` helper that we can aim to improve in the future, since the very small change "just works" for many users (and especially most default GHA runners).

PR #622, opened 2025-11-11
Fix rare count_test edge case (#2)

maksymiuks:
Hi!

In our enterprise solution, we have discovered a peculiar edge case when `count_test` function is called with `.current_test` not yet initialized, leading to error:

```
Error in tests$.data[[1L]] <- .current_test$index : 
  replacement has length zero
```

For us, we encountered it when calculating coverage for the  `digest` when using `tinytest` Ubuntu libraries (funnily enough, when installing tinytest from source, the issue did not appear). That makes reproducing the issue really tedious, although I was able to do it on macOS as well. Nevertheless, I believe that there is no harm in adding that if statement to cover that edge case, even if it's super rare, simply to make sure the `covr` is not vulnerable to weird/unexpected package combinations in the future.

### DIFF
--- a/R/trace_tests.R
+++ b/R/trace_tests.R
@@ -114,6 +114,9 @@ count_test <- function(key) {
     tests$tally <- rbind(tests$tally, matrix(NA_integer_, ncol = 4L, nrow = n))
   }
 
+  # ignore if .current_test was not initialized properly yet
+  if (length(.current_test$index) == 0) return()
+
   # test number
   tests$.data[[1L]] <- .current_test$index
 
--- a/R/trace_tests.R
+++ b/R/trace_tests.R
@@ -115,7 +115,9 @@ count_test <- function(key) {
   }
 
   # ignore if .current_test was not initialized properly yet
-  if (length(.current_test$index) == 0) return()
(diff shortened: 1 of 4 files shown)


R/trace_tests.R
```diff
@@ -114,6 +114,9 @@ count_test <- function(key) {
     tests$tally <- rbind(tests$tally, matrix(NA_integer_, ncol = 4L, nrow = n))
   }
 
+  # ignore if .current_test was not initialized properly yet
+  if (length(.current_test$index) == 0) return()
```
jimhester:
```suggestion
  if (length(.current_test$index) == 0) {
    return()
  }
```

jimhester (commented):
Had one minor style change suggestion.

Also can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

maksymiuks:
@jimhester, I updated the PR to include another edge-case fix. I also added an issue for both these problems so they can be linked in NEWS. Could you please take a look at it?

jimhester:
Sounds good, thanks for working on this!

PR #632, opened 2026-01-07
Account for vectors of length 0 when cleaning coverage

### LINKED ISSUE #631: Rare edge cases in `count_test` and `clean_coverage_tests`
Hi!

In our enterprise solution, we have discovered a peculiar edge case when `count_test` function is called with `.current_test` not yet initialized, leading to error:

```
Error in tests$.data[[1L]] <- .current_test$index : 
  replacement has length zero
```

For us, we encountered it when calculating coverage for the  `digest` when using `tinytest` Ubuntu libraries (funnily enough, when installing tinytest from source, the issue did not appear). That makes reproducing the issue really tedious, although I was able to do it on macOS as well. Nevertheless, I believe the most straightforward approach is to add an if statement to cover that edge case, even if it's super rare, simply to make sure the `covr` is not vulnerable to weird/unexpected package combinations in the future.

Similarly, we've experienced another rare edge case when testing the full `renv` testing suite:

```
```
Error in `if (is.null(n <- nrow(obj[[i]]$tests$tally)) || n <
   val) next`:
   ! missing value where TRUE/FALSE needed
```

As in the previous case, I believe that making sure that the value there is always logical does not harm the package, but ensures its stability even in weird edge cases.

### PR DESCRIPTION
**Pull request by maksymiuks**
Hi @jimhester, I'm sorry for bothering again. Apparently, my previous PR didn't cover an edge case of an edge case, or rather, by solving one, I created another. I'm sorry for that.

https://github.com/r-lib/covr/blob/f1866d296c00884d1f085ff245669de01bc864c4/R/covr.R#L669

`val` variable, in particular scenarios,  can have `integer(0)` value and because of this R behaviour:

```
> FALSE || logical(0)
[1] NA
```
we end up with an NA in the conditional. That's why I've rewritten the condition to use `length(val) == 0` instead of `is.null(val)`. It is more robust and covers more of the cases that are problematic for that function, because both `NULL` and `integer(0)` are of length 0. There should be no more problems with that function anymore. 

Closes #631


jimhester:
Could you add a test case to verify we don't break this in the future, thanks!

maksymiuks:
@jimhester Added the test. I figured it's better to prepare a dummy problematic object rather than an example package, so the test case is more robust and we are sure the function handles the problem well regardless of the source.

